PR #18, opened 2016-01-04
Added support for testing Arch Linux.

andy-maier:


andy-maier:
This merge request will be superseded by a new one, if we agree on the proposal described in issue #20.

Issue #20, opened 2016-01-04
Overall distro test missing


andy-maier:
The unit test tests the distro release file, the OS release file, and the lsb_release command independently from each other.

However, the ld.info() method and the underlying specific methods depend on these three sources of information coming together in a particular combination, for each distribution. For example, the precedence rules implemented by these functions are currently untested.

I suggest to add a class for testing all the combinations, for the distros that are in test scope.

I intend to come up with a pull request for that, shortly.

andy-maier:
It seems to me that the data structure in the tests/resources directory needs to be changed in order to accomplish this. For example, the current files that have release numbers in their file names don't work, because part of the algorithm is to look at the file name of the distro files (and they are not used in the tests currently). Also, part of the algorithm is to list files according to a search pattern and to process the list in a sorted order (the algorithm also includes a lookup in the DIST_TO_BASE which is why the test still works, but I think this places too much knowledge about the implementation into the test).

My proposal for a new file structure is:

    +-- resources
        +-- distros
        |   +-- ubuntu14
        |   |   +-- bin
        |   |       +-- lsb_release
        |   |   +-- etc
        |   |       +-- os-release
        |   |       +-- ubuntu-release
        |   |       +-- lsb-release
        |   + ...
        +-- special
            +-- empty-release
            +...    # and other corner cases

The files only exist if they exist in the real distro, and if the real distro has symbolic links, that is represented as well in the file structure (for example, CentOS 7 has a /etc/centos-release file and symbolic links /etc/system-release and /etc/redhat-release that point with a relative file name to /etc/centos-release. In some distros, /etc/os-release is a symbolic link to /usr/lib/os-release, that would be represented as well, and it is important to test whether symbolic links are found when searching for the file pattern.

The bin/lsb_release command is a shell script that exists in two versions: one with (like Ubuntu 14) and one without (like CentOS 7) the <distro>/etc/lsb-release file. Both versions only support the -a option. I have written both versions already, and these shell scripts are both quite short.

Alternatively, we could have one version of the lsb_release script that uses an lsb-release file and we vary the location of the lsb-release file between <distro>/etc and <distro>/hidden, depending on whether the distro has it in /etc or not.

The reason to be so picky about the presence of the lsb-release file is that it matches the file search pattern that is used, and sorts before or after the distro release file, depending on the distro name. Again, the check for DIST_TO_BASE makes sure it is never used when the distro release file is being identified, but again I think depending on that circumstance in the test code is too much of a white box.

[...]

PR #25, opened 2016-01-06
New file structure in tests/resources for overall distro test (issue #20)

andy-maier:
This merge request contains the changes for the changed file structure.

I have added release files for some distros and the existing test approach has now test code for all existing distro files. These tests run again fine on the changed file structure.

I also made sure that the release files are exactly like they are on real distros, including thi8ngs like trailing newlines. Therefore, this merge request has a (runtime) dependency on merge request #24 (without it, the unit tests fail).

The code for the overall distro test has not yet been written, I wanted to do this in two steps.

The change in .gitignore was necessary because git apparently applies directories to the whole subtree, so the entries "bin/" and "lib/" caused the tests/resources/distros/<distro>/bin and tests/resources/distros/<distro>/usr/lib directories to be excluded. Adding a "/" in front of such directories fixes that. I verified that bin and lib directories in the top level directory are still excluded. I did this not only for the two directories that needed it, but for all that I think are not intended to be excluded in the whole subtree. This is worth reviewing, of course.

Please review and merge, if you agree with the new file structure.


andy-maier:
The reason the Travis CI tests fail is because of the runtime dependency on the merge request #24.

Issue #26, opened 2016-01-11
Exception when lsb_release returns no LSB information


andy-maier:
When the lsb_release command is present, and when it returns no LSB conformance information, its first line usually (i.e. on real distributions) reads: 

    No LSB modules are available.

Because this does not contain any colon, the following code in _parse_lsb_release() raises an exception about not enough items in the returned tuple:

        props = {}
        for obj in content:
            k,v = obj.strip('\n').split(':', 1)         # raises exception when no colon in obj
            props.update({k.replace(' ', '_').lower(): v.strip()})
        return props

The following code addresses that:

        props = {}
        for obj in content:
            kv = obj.strip('\n').split(':', 1)
            if len(kv) != 2:
                # Ignore lines without colon.
                continue
            k,v = kv
            props.update({k.replace(' ', '_').lower(): v.strip()})
        return props

nir0s:
lgtm

PR #28, opened 2016-01-11
Fix for issue #20: Overall distro test missing, part 2

andy-maier:
This merge request contains the implementation of the overall distro test, based on the new file structure under tests/resources.

In addition, it contains two fixes for issues that came up during the overall distro test:

1. `_attempt_to_get_release_file()` now returns the file path, instead of just the base file name. The prior code caused `distro_release_info()` to not find the file.

2. Not a real problem, but it helps when looking at an ldi object: Moved the determination of the distro file name using `_attempt_to_get_release_file()` into `__init__()`, so that the file that was found is stored in `self.distro_release_file`. In the prior code, the location of the file that was read was never stored anywhere in the ldi object, making debugging harder than needed. The `__init__()` method (which is called during module load because of the global `_ldi` object) already performs file access, so I thought one more test for a file does not make it much worse. But it is a review point: If we are concerned about file access at module load time, we should defer all file access to the first use of some data, not just this one.

In addition, fix26 (see issue #26) is the base for this merge request, so its commit is included here, because the overall tests depend on it.

Please review and merge.

### DIFF
--- a/ld/__init__.py
+++ b/ld/__init__.py
@@ -11,7 +11,8 @@ class LinuxDistribution(object):
                  os_release_file='',
                  distro_release_file=''):
         self.os_release_file = os_release_file or const._OS_RELEASE
-        self.distro_release_file = distro_release_file or ''
+        self.distro_release_file = distro_release_file or \
+            self._attempt_to_get_release_file()
         self._os_release_info = self.os_release_info()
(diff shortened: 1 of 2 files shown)


nir0s:
please fix conflicts. for all other PRs as well.
Thanks

andy-maier:
I resolved the conflicts, and commit fdaac70 is fine, now.

PR #30, opened 2016-01-12
Added testcase for Mageia 5; Made parsing of distro release file tolerant against trailing stuff; Added repr() in LDI class.

andy-maier:
This merge request adds a testcase for the Mageia distro, release 5. This distro has a few interesting properties:

* It has a file `/etc/release` (a symlink to `/etc/mageia-release`)
* It has a file `/etc/version` which contains a more precise version than the distro release file (this file is not being used right now, but we could add support for obtaining the most precise version in the future)
* Its distro release file has the text ` for x86_64` after the `(codename)` part. This required extending the parsing pattern for the distro release file such that it now tolerates (and ignores) such trailing text.
* Its `lsb_release -a` command returns `LSB Version:   *` which is a good additional test for parsing its output.

Note that I have added `'mageia': 'mandrake'` to DIST_TO_BASE.

It also includes the previous two merge requests #27 amd #28, because it depends on them.


andy-maier:
The Travis test fails because I had a bug in the lsb_release script that only had an impact when the test runs on Ubuntu (it treats a path of "/../etc/lsb-release" as a valid path and strips off the ".." part). Will fix and create a new merge request.

andy-maier:
Resolved the merge conflicts, and commit 871b830 is fine, now.
The merge request includes the commit for "Fix for issue #20 (try4)..." because it depends on it.

Issue #48, opened 2016-01-26
Ubuntu returns distro release file info as 'debian'


andy-maier:
**Problem:**

On Ubuntu 14.04, the information from distro release files returns:

    ld._ldi.dist = "debian"
    ld._ldi.distro_release_info() = {'codename': '', 'name': '', 'version_id': ''}

This is caused by using the `/etc/debian_version` file as the distro release file, because that matches the search pattern for distro release files, and 'debian' is found in `DIST_TO_BASE`.

When using the consolidated result, the os-release information supersedes the distro release file information, but if a user uses `distro_release_info()` or `get_distro_release_attr()`, this discrepancy will surface.

**Proposed solution:**

All of the following:

1. Remove `*version` from the file search pattern for distro release files. Out of all distros that I know of having such a file (Ubuntu, Debian, Mageia, Slackware), none of these files follows the `<name> release <release> (<codename>)` pattern.

2. In function `_attempt_to_get_release_file()` which determines the file name of the default distro release file to be used, make the decision to use a file not only based upon presence of its first part of the file name in the `DIST_TO_BASE` dict, but in addition based upon whether it provides at least some information when parsing it with the usual pattern. Ideally, this should be done such that the parsing is performed only once.

This approach leaves the possibility to add support for parsing `*version` files with their specific format in the future. It just removes today's inconsistency of searching for these files but then not understanding their format.

andy-maier:
Solved with PR #49. I am closing this issue.

Issue #40, opened 2016-01-20
Information from unknown distros is ignored


andy-maier:
**Problem:**
If a distro has just a distro release file (/etc/*-release) but no /etc/os-release file and no lsb_release command, and if the first part of the file is not one that is known to ld, then the information in that file is ignored, and all ld functions return the empty string.

**Solution:**
I think that should be changed to just accept what such a file provides. That is better than ignoring it.

I think the reason for the filtering to known distros is that on distros that have multiple distro release files (RHEL, Mageia, many others), this is the way how the one to be used is found. But the default case for not finding a known distro could be changed to pick the first file after excluding a few generic ones such as system-release.

nir0s:
So you're suggestion that we look for the first *-release file which isn't a generic one and parse it? That should at least return something allows the user to know that ld did not identify it. Consider a case where it found some arbitrary *-release file that a weird distro decided on or a user decided to inject.. and suddenly you get some weird info there.

andy-maier:
This new approach would only be used for unknown distros (that do no not show up in _DIST_TO_BASE). So if a user injects etc/*-release files into a known distro, that would not change the result.

For unknown distros, a user injecting new /etc/*-release files would cause the result to be changed, that is true. But if that user edits a release file that is used by ld (even for a known distro), the result also gets changed. You cannot project ld from users that are tampering with the distro.

The only problematic case I can see is when a distro not known to ld has some generic release file that ld does not yet have on its ignore list. At that point, we can deal with the issue if users bring it up, either by adding the new generic release file name to ld's ignore list, or by making the new distro one that is known to ld.

andy-maier:
Solved with PR #49. I am closing this issue.

PR #49, opened 2016-01-26
Fixes for issues #40 and #48

andy-maier:
This PR contains fixes for issues #40 and #48, as proposed in the descriptions.

This includes some restructuring in the processing of distro release files.

Also, the distro release file dict now contains only those entries that were found (as opposed to all
entries with the empty string if no information was found). The users of accessor functions like `id()` or `get*attr()` see no change due to this, because they default non-existing entries to the empty string.

Finally, the distro ID found in the file name of the distro release file is no longer stored in `self.dist`, but as key 'id' in the distro release file dict. This seems to be more consistent, to me.

Note the change in expected test case results, when reviewing.

### DIFF
--- a/ld/__init__.py
+++ b/ld/__init__.py
@@ -11,8 +11,7 @@ class LinuxDistribution(object):
                  os_release_file='',
                  distro_release_file=''):
         self.os_release_file = os_release_file or const._OS_RELEASE
-        self.distro_release_file = distro_release_file or \
-            self._attempt_to_get_release_file()
+        self.distro_release_file = distro_release_file or ''
         self._os_release_info = self._get_os_release_info()
         self._lsb_release_info = self._get_lsb_release_info() \
             if include_lsb else {}
@@ -23,13 +22,11 @@ class LinuxDistribution(object):
             "LinuxDistribution(" \
             "os_release_file=%r, " \
             "distro_release_file=%r, " \
-            "dist=%r, " \
             "_os_release_info=%r, " \
             "_lsb_release_info=%r, " \
             "_distro_release_info=%r)" % \
             (self.os_release_file,
              self.distro_release_file,
-             getattr(self, "dist", None),
              self._os_release_info,
              self._lsb_release_info,
              self._distro_release_info)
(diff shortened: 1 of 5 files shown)


ld/__init__.py
```diff
@@ -69,27 +66,81 @@ def _get_lsb_release_info(self):
...
+            # If it was specified, we use it and parse what we can, even if
+            # its file name or content does not match the expected pattern.
+            distro_info = self._parse_distro_release_file(
+                self.distro_release_file)
+            basefile = os.path.basename(self.distro_release_file)
+            # The file name pattern for user-specified distro release files
+            # is somewhat more tolerant (compared to when searching for the
+            # file), because we want to use what was specified as best as
+            # possible.
+            release_file_pattern = re.compile(r'(\w+)[-_](release|version)')
```
nir0s: can we move this to globals?

ld/__init__.py
```diff
@@ -69,27 +66,81 @@ def _get_lsb_release_info(self):
...
+                # TODO: Normalize distro_id
+                distro_info['id'] = distro_id
+            return distro_info
+        else:
+            files = os.listdir(const._UNIXCONFDIR)
+            # We sort for repeatability in cases where there are multiple
+            # distro specific files; e.g. CentOS, Oracle, Enterprise all
+            # containing `redhat-release` on top of their own.
+            files.sort()
+            release_file_pattern = re.compile(r'(\w+)[-_]release')
```
nir0s: and then use it here as well?

andy-maier:
I assume you are talking about the `release_file_pattern` variable?

Of course they could become globals, but the reason I removed 'version' from the pattern used for searching the distro release file if not specified, is issue #48. if we think there are distro release files that end with 'version' and where there is no other release information, then we could also have 'version' in the pattern but special-case the `debian_version` case on Ubuntu. However I'd prefer not having 'version' and to wait for people finding a distro that needs support for 'version'.

I just added 'version' again this morning in a private version to double check this, and was wondering why the TestDistRelease did not surface the error of issue #48, and it turns out the reason is that this class does not have a testcase for Ubuntu. I will add one and will update the PR.

I don't see a particular reason for having 'version' in the pattern used for the case where the distro release file was specified by the user. I just did not remove it, because there was no issue similar to #48. I would not object to removing 'version' from this pattern as well.

For now, I'll update the PR with the Ubuntu test case, and I wait for your opinion on making the variables global, and about the use of 'version' in their patterns.

andy-maier:
This time, I left the commits separate because the one that reordered the test methods would have convoluted the other changes quite a bit.

On commit eb15b3a:
- I found that adding an Ubuntu testcase for the distro release files does not make sense, because debian_version in ubuntu14 should not really be considered a distro release file. Nevertheless, I have added that file for better testing.
- Because the `slackware-version` file can be considered a distro release file, I added `*version` again to the pattern when the distro release file is searched, and moved the resulting single pattern definition into a global variable. I also moved the pattern for the distro release file content into a global variable.
- I found that I had forgotten to add the support for ignoring certain generic or otherwise excluded distro release file names, and added that now.
The commit message details the changes.

PR #51, opened 2016-02-05
Fixes #36: Unreliable id() and meaning of base().

### LINKED ISSUE #36: Unreliable id() and meaning of base()
There is not a reliable distro ID string currently. The result of the id() method is not always reliable, and the result of the base() method is somewhat more reliable, but its meaning is not entirely clear.

Example for CentOS/RHEL/Fedora/Oracle (return values shown have been taken from the TestOverall test cases, which have been verified to represent the situation in the real distros, so 
this should represent the actual behavior on these distros):

    Distro         id()      base()
    RHEL 6.5       redhat    rhel
    RHEL 7.0       rhel      fedora
    Fedora 23      fedora    fedora
    CentOS 5.11    centos    rhel
    CentOS 7.0     centos    rhel
    Oracle 7.1     oracle    rhel

Issue: The id() value between RHEL 6.5 and RHEL 7.0 changes.
Issue: CentOS is the open source distro of RHEL, but is it "based on" RHEL, or not rather based on Fedora?
Issue: Is Oracle really based upon RHEL in the sense that RHEL's source code is the input for Oracle, or is it not rather using CentOS's source code (I did not research it, but using RHEL does not sound plausible to me).

It seems to me that what we need is:

1) a more reliable id() value, that is normalized to a certain degree, e.g. by a mechanism similar to the DIST_TO_BASE dictionary, but in such a way that Oracle, RHEL, Fedora, CentOS would remain separate strings, and the string for each distro would not suddenly change in a new release, just because the distro moves to a new type of release file (RHEL 7.0 introduced /etc/os-release, which caused this change).

2) a better semantic definition on what base() really means. For example, does it denote:
   a) where the source code comes from (RHEL comes from Fedora, CentOS comes from RHEL, Oracle comes from xxx)?
   b) some sort of family (Fedora, RHEL, CentOS, Oracle would all be the same family)?
   c) something similar to like()? Note that the implementation of base() uses the result of like() for distros that are not in DIST_TO_BASE, but like() is using the LIKE attribute in /etc/os-release which sometimes contains more than one distro ID. If we stick with using like() as the default, then the description of base() should mention that it may end up returning more than one string.

Comments / Opinions?

andy-maier:
This PR fixes issue #36: Unreliable id() and meaning of base().
Please review and merge.
In addition to what the commit message says, I have also updated the docstrings for those methods that had to do with the change.

Details from the commit message:
- Added support for a reliable distro ID returned by `id()`.
  The normalization happens only for `id()`; the `*_release_info()` and
  `get_*_release_attr()` methods do not provide a reliable ID; these
  methods provide the values as obtained from the respective sources.
- The set of distros for which the ID is reliable has been initially
  taken from the `_DIST_TO_BASE` dict. A TODO has been added to make
  sure we review that list.
- These reliable distros are documented with their ID values in the
  description of the `id()` method.
- Removed the `base()` method and, because it was no longer used, its
  `_DIST_TO_BASE` dict.
- Adjusted testcases and standalone query script to the changed behavior
  of the `id()` method.
- Documented a TODO in the `TestOverall` test class for those distros that
  have reliable IDs, but have no testcase representation so far.
- Changed the mocking concept for the `TestLSBRelease` class to also use
  the concept from `DistroTestCase`. This became necessary because the
  previous mocking concept was making assumptions about the
  implementation that were no longer met. As part of that, the LSB
  output files in the "special" directory became simple `lsb_release`
  scripts that print a hard coded set of lines, in the "testdistros"
  directory.


andy-maier:
Comments?

PR #52, opened 2016-02-08
Fixes #29: Version sometimes looses precision.

### LINKED ISSUE #29: Version sometimes looses precision
**Problem:**
I have seen one case where the defined priority of first os-release, then lsb_release, then distro release file, causes version() to return just the major version, when the system actually also has a minor version available. So we do not return all possible version detail, i.e. we "loose some precision".

The case I have seen is Debian 8.2, and that case is actually represented in the new test structure exactly as I have seen it on a real Debian 8.2 system:

  * /etc/os-release file defining VERSION=8, VERSION_ID=8
  * /etc/debian-version file containing just "8.2"

The current code simply goes by priority.

**Solution:**
If the code would be changed to go by best precision (out of all available sources), this could be addressed. The downside of such a solution is that it makes inconsistencies between the multiple sources even harder to understand, particularly if the inconsistency is not just more or less precision but even contradictions in the version parts that are available in multiple sources).

I think however, in the end it would be good to use the highest available precision for the version.

andy-maier:
Please review.
This PR is largely independent of PR #51 (there is a trivial merge conflict, which I can resolve either way, whichever you pick first).

There are several things to watch out for when reviewing:
- It did not become a new `best_version()` method as suggested in a prior comment on the discussion of issue #29, but a new `best` parameter on the existing version related methods, that by default works as today, and that can be used to turn on the best precision algorithm. This is because there were five methods related to versions, and it did not seem appropriate to me to double their number.
- I first thought we need to add more sources for the version, i.e. the description fields of `os-release` and `lsb_release`, so I implemented support for them, by using their version if their string matches the format of the one-line distro release files. However, it turns out that there is no case in the currently tested distributions where the version in these strings is more precise than the most precise version available in the official version related fields. In other words, the job to get the best precision can be achieved by just looking at the official version fields, without these new sources. So the question is do we remove these new sources again, or have them just in the best precision mode, or in both modes (current code of the PR).
- `info()` uses the fixed priority approach for its versions, with no possibility currently to select the best precision approach. Should a switch be added, or should it always be most precise, or always use the fixed priority version (current code of the PR)?

As always, I'm willing to change the implementation as needed, so don't feel obliged to just accept the PR.

More details, from the commit message:
- Added a `best` parameter to `version()`, `version_parts()`, `major_version()`, `minor_version()` and `build_number()`.
- Added the `PRETTY_NAME` of `os-release` and the `Description` of `lsb_release` as additional sources for parsing them in the distro release file format for a version. This needs review!
- Added a global `version_parts()` function that was probably forgotten.
- Fixed the potential for an `AttributeError` in `version_parts()` that occurs if the major version does not just consist of digits. Now, an empty string is returned in that case. It is admittedly an unlikely case, but handling it does not cost much.
- Added descriptions for version-related methods of `LinuxDistribution`.
- Added testcases for `best=True`.

### DIFF
--- a/ld/__init__.py
+++ b/ld/__init__.py
@@ -422,37 +422,117 @@ class LinuxDistribution(object):
                     name = name + ' ' + version
         return name or ''
 
-    def version(self, pretty=False):
-        """Returns the version of a specific distribution.
+    def version(self, pretty=False, best=False):
+        """Returns the version of the distribution as a string.
 
-        If pretty=False, the version is returned without codename (e.g. 7.0).
-        If pretty=True, codename is appended (e.g. 7.0 (Maipo))
+        If `pretty=False`, the version number is returned without codename
+        (e.g. "7.0").
+        If `pretty=True`, the codename in parenthesis is appended (e.g.
+        "7.0 (Maipo)"), if the codename is non-empty.
+
+        Some distributions provide version numbers with different precisions in
+        the different sources of distribution information. Examining the
+        different sources in a fixed priority order does not always yield the
+        most precise version (e.g. for Debian 8.2, or CentOS 7.1).
+        The `best` parameter can be used to control the approach for the
+        returned version:
+        If `best=False`, the first non-empty version number in priority order
+        of the examined sources is returned.
+        If `best=True`, the most precise version number out of all examined
+        sources is returned.
+
+        Details:
+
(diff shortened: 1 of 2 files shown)


andy-maier:
Comments?

andy-maier:
Rebased on master and resolved a trivial conflict. Travis says ok. Merged.

PR #53, opened 2016-02-08
Fixes #50: Rules for processing values in OS release file…

### LINKED ISSUE #50: Rules for processing values in OS release file not obeyed
The definition of the format of OS release files at http://www.freedesktop.org/software/systemd/man/os-release.html has several rules that are not currently implemented in ld:

Quoting from that web page:

> The basic file format of os-release is a newline-separated list of environment-like shell-compatible variable assignments. It is possible to source the configuration from shell scripts, however, beyond mere variable assignments, no shell features are supported (this means variable expansion is explicitly not supported), allowing applications to read the file without implementing a shell compatible execution engine. Variable assignment values must be enclosed in double or single quotes if they include spaces, semicolons or other special characters outside of A–Z, a–z, 0–9. Shell special characters ("$", quotes, backslash, backtick) must be escaped with backslashes, following shell style. All strings should be in UTF-8 format, and non-printable characters should not be used. It is not supported to concatenate multiple individually quoted strings. Lines beginning with "#" shall be ignored as comments.

The following of these rules are not currently implemented when parsing the file:
* Unescaping of backslash-escaped special characters ("$", quotes, backslash, backtick). The current code does not treat backslashes in any special way and thus treats them literally.
* Values are UTF-8 encoded. The current code assumes ASCII encoding.
* Strip only one level of surrounding quotes. The current code removes any single or double quotes, even inside the string.
* Ignore lines starting with "#". The current code uses such a line if it contains a "=".
* Allow string values that extend over multiple lines (The text on the web page allows that, see "shell-compatible"). The current code works strictly line by line and assumes each variable is defined on a single line.

Solution: Implement these rules.

andy-maier:
This PR fixes issue #50. Please review and merge.
I think it can be merged independently of the other open PRs; if not I'll rebase them.

Details from the commit message:
- used the `shlex` module to do the shell-like parsing of `os-release` files.
- added unit testcases for the `_parse_key_value_files()` method (class `TestKeyValueParsing`).

### DIFF
--- a/ld/__init__.py
+++ b/ld/__init__.py
@@ -1,6 +1,8 @@
+import sys
 import os
 import re
 import subprocess
+import shlex
 
 from . import constants as const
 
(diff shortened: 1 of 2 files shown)


andy-maier:
Comments?

nir0s:
Andy, I have some issues that don't allow me to progress fast enough with ld. Added you as collaborator. You're invited to continue. I'll try to respond as much as possible.

andy-maier:
Nir, thanks much for trusting me - I appreciate that!
Do you want some more time to review the three PRs that are up, or would you be ok with me merging them now?

nir0s:
Unfortunately, I can't find the time for it right now. I'll be back
cracking on this in a month or so. You can merge as you see fit.

Thanks!

Issue #54, opened 2016-02-18
Test: Improved testcases for info() that test build_number


andy-maier:
In the implementation of `LinuxDistribution.info()`, the returned dictionary is built from the data in the `self` object, except for `build_number` which is taken from the module-global `_ldi` object. This may result in a discrepancy if the user creates a `LinuxDistribution` object on their own that uses different input files, like our unit test cases do for example.

The reason this did not come up yet in our unit tests is that we do not have a test with a distro that has more than two parts of the version number.

**Solution:**
* Add a unit test case (real or artificial) that has all three parts in its version number.
* In `info()`, invoke `self.build_number()` instead of `build_number()`.

andy-maier:
This issue was closed because PR #56 by accident contained commit 33bfbcd. That commit fixes the ld code, but it does not contain the extension for the testcases that would have caught this issue.

Reopening this issue, in order to provide the extended testcases.

andy-maier:
PR #113 adds a comprehensive test for info() that compares every field in its result with the result of the corresponding access function, for all distros. That test would have caught this issue.
This issue is therefore fully addressed now and can be closed.

PR #56, opened 2016-03-02
Query script hashbang

andy-maier:
PR #51 added a hashbang to the `query_local_distro.py` script, but did it incorrectly (and apparently without any testing ... :-(

This PR fixes the hashbang.

I verified that no other Python script contains an incorrect hashbang.


andy-maier:
This PR by accident included the fix for issue #54, as far as the code is concerned. The extension of the testcases was not yet in that commit; they will be handled as part of issue #54.

PR #104, opened 2016-04-13
103 replace name

nir0s:


nir0s:
Note that I'm not keeping backward compatibility and `ld` is not available anymore. We can, if we want, still instantiate an `ld` object. Should we maintain backward compatibility or break things now before this is implemented in different places?

andy-maier:
+1 from me. Sorry for the delay.

nir0s:
I'll leave this PR opened until we decide which name we want to go with.

PR #111, opened 2016-04-20
remove get_ prefix from attribute functions

nir0s:


andy-maier:
Reviewed the code.
+1

PR #113, opened 2016-04-20
Fixes #94, fixes #112: Improved and fixed info() function.

### LINKED ISSUE #94: Should info() use first or best version for its version-related results?
Currently, in its result, the `info()` method provides the first available version of the (possibly multiple) underlying data sources (i.e. specify argument `best=False`).

It would be possible to use the version with the best precision, instead (i.e. specify argument `best=True`).

Please provide your review/considerations/wishes about which way to go in this matter, by commenting on this issue.

### LINKED ISSUE #112: info() result does not include codename
The documentation for `info()` shows a `codename` key in its result dictionary.
The code for `info()` does not produce that key.
The testcases for `info()` so far did not catch that discrepancy.

Proposal:
* Change `info()` implementation to include a `codename` member in its result.
* Improve the `info()` testcases to verify that.

andy-maier:
Ready to be merged.
Please review.

Details from the commit log:

- Added a `best` parameter to `info()`.
- Fixed implementation of `info()` by adding the missing `'codename'` item in its result.
- Improved test cases for `info()` by adding a generic all-distro test that compares the result of `info()` with the result of the specific consolidated accessor functions.
- Minor editorial improvements in documentation of `info()`.
- Removed todo3 about issue #94.

### DIFF
--- a/tests/test_ld.py
+++ b/tests/test_ld.py
@@ -1046,7 +1046,7 @@ class TestGetAttr(DistroTestCase):
             for key in info.keys():
                 self.assertEqual(info[key],
                                  ldi.os_release_attr(key),
-                                 "distro: %s, key: %s" % (distro, key))
+                                 "distro: {}, key: {}".format(distro, key))
 
     def test_lsb_release_attr(self):
         distros = os.listdir(DISTROS)
@@ -1059,7 +1059,7 @@ class TestGetAttr(DistroTestCase):
             for key in info.keys():
                 self.assertEqual(info[key],
                                  ldi.lsb_release_attr(key),
-                                 "distro: %s, key: %s" % (distro, key))
+                                 "distro: {}, key: {}".format(distro, key))
 
     def test_distro_release_attr(self):
         distros = os.listdir(DISTROS)
@@ -1072,7 +1072,7 @@ class TestGetAttr(DistroTestCase):
             for key in info.keys():
                 self.assertEqual(info[key],
                                  ldi.distro_release_attr(key),
-                                 "distro: %s, key: %s" % (distro, key))
+                                 "distro: {}, key: {}".format(distro, key))
 
 
 class TestInfo(DistroTestCase):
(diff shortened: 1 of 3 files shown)


nir0s:
Please fix conflicts.

andy-maier:
Fixed.

tests/test_ld.py
```diff
@@ -1115,6 +1131,71 @@ def test_linux_disribution_full_false(self):
...
+        for distro in distros:
+            self._setup_for_distro(os.path.join(DISTROS, distro))
+
+            ldi = ld.LinuxDistribution()
+
+            info = ldi.info()
+
+            self.assertEqual(info['id'],
+                             ldi.id(),
+                             "distro: %s" % distro)
```
nir0s: Is there a reason to use % string formatting here instead of curl braces?

andy-maier:
Not that I know of, other than the fact that it was used in `ld` already, and maybe that I'm simply used to it. I have no problem moving to the new `string.format()` syntax, but there is more to change then than just this one place.

andy-maier:
The discussion in issue #112 suggests to also support `pretty`in `info()`.
Plus, the discussion above suggests to start using the new string formatting language. (... as new as Python 2.6 ;-)

andy-maier:
The last commit does all that, and updates the testcases accordingly.
Please re-review.

nir0s:
tests are broken

andy-maier:
The issue in the last commit was that support for the `{}` syntax without field name or number was introduced only in python 2.7, and my local test machine does not have py26 installed, so i did not realize it upfront.

tests/test_ld.py
```diff
@@ -1046,7 +1046,7 @@ def test_os_release_attr(self):
             for key in info.keys():
                 self.assertEqual(info[key],
                                  ldi.os_release_attr(key),
-                                 "distro: %s, key: %s" % (distro, key))
+                                 "distro: {}, key: {}".format(distro, key))
```
nir0s: this won't pass in 2.6. Please use {0} ...

PR #123, opened 2016-04-23
Rearrange test env

nir0s:
This is a complete refactoring of the testing env.


nir0s:
For the reviewer: note that there are still some things to do there but I'd like to first merge this so that we have a good base.

funkyfuture:
> so that we have a good base.

indeed it is.

andy-maier:
+1

Issue #74, opened 2016-03-16
Testcase for Amazon Linux is missing


andy-maier:
Amazon Linux (ID: `amazon`) is one of the distros with a reliable ID. A testcase should be added.

If you want to help, it is sufficient to post to this issue:
* The content of the `/etc/os-release` file, if any.
* The file names and content of the `/etc/*release` and `/etc/*version` files, if any.
* The output of the command: `lsb_release -a`, if available.

nir0s:
The Amazon Linux 2016.03 image contains an os-release file while the 2014.03 one only contains a system-release file, which we do not handle. Before this is merged, I suggest we discuss this.

nir0s:
Will open an issue regarding system-release files in general

PR #125, opened 2016-04-30
add amazon linux test case

nir0s:
Added test case for #74 (Amazon Linux).


andy-maier:
+1
This is the first time the exclusion of system-release hurts. One option to address this would be to exclude it only if there are other *-release files.

nir0s:
True. Will open an issue.

Issue #124, opened 2016-04-29
Importing distro on windows/mac results in unexpected behavior


nir0s:
Trying to import distro on Windows/Mac results in some weird behavior. For instance, on Windows, it tries to execute `lsb_release -a` and exists on a ProcessError.

I suggest we test that:

```python
if not sys.platform.startswith('linux'):
    raise ImportError('distro currently only supports Linux')
```

That way, users can do:

```python
try:
    import distro
except ImportError:
    pass
```

which seems reasonable to me.

Thoughts?

MartijnBraam:
I think it might be beter to return a generic "Unsupported distribution / operating system" distro name when the distro cannot be detected.

I assume this is also a problem on linux distributions without any release information files.

nir0s:
Linux distros with no info return empty strings.

MartijnBraam:
I think that raising an ImportError is not the right error and not the right place to raise an exception for this. 

It is a little hard to change that since the linux distro is detected as soon as the module is imported. Maybe also return empty strings for osx and windows and have a `distro.is_linux` property. 

Another option is simply having the id `windows`, `osx` and `genericlinux` for when no information is available

nir0s:
I like the empty strings with the distro.is_linux property suggestion.

nir0s:
I tried implementing this.. and it seems to create more confusing and overhead than just raising an ImportError exception. `sh` does the same thing (https://github.com/amoffat/sh/blob/master/sh.py#L34)

I believe we should go with this.

nir0s:
PR open. Let's keep discussing.

PR #129, opened 2016-05-02
raise exception when importing on non-supported platforms

nir0s:
* Added Appveyor testing to verify that importing distro fails on Windows.
* Now skipping all tests if not on Linux.
* Added ImportError Exception when importing from non-linux platforms.


nir0s:
Resolves #124

andy-maier:
+1

PR #130, opened 2016-05-14
Manjaro Linux

funkyfuture:
this adds full support for Manjaro Linux (generically) and includes changes that i found useful on the way to achieve that.


funkyfuture:
@nir0s can you give me hint what `appveyor` is and why it might be failing?

nir0s:
@funkyfuture, thanks for contributing. Appveyor is travis for windows. I added it to always make sure that you can't import distro on non-linux machines. It fails because the PR to support this hasn't been merged yet. I configured appveyor to ignore branches without appveyor.yml files so you're good to go.

distro.py
```diff
@@ -1034,6 +1041,8 @@ def _parse_distro_release_content(line):
                 distro_info['version_id'] = m.group(2)[::-1]
             if m.group(1):
                 distro_info['codename'] = m.group(1)[::-1]
+        elif line:
+            distro_info['name'] = line.strip()
```
nir0s: I don't quite understand this...

setup.py
```diff
@@ -39,7 +39,7 @@ def read(*parts):
     description='Linux Distribution - a Linux OS platform information API',
     long_description=read('README.rst'),
     py_modules=['distro'],
-    install_requires=['six'],
+    install_requires=[],
```
nir0s: I'm glad you removed the dependency. you can also just remove the `install_requires` argument then.

tests/resources/distros/__shared__/bin/lsb_release
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
nir0s: wonderful. thanks

tests/test_ld.py
```diff
@@ -26,15 +26,18 @@
 import distro
```
nir0s: #123 implements a whole refactoring of the test env. mind basing your tests on that branch?

distro.py
funkyfuture: the content of [manjaro-release](https://github.com/nir0s/distro/pull/130/files/94ef753bc172803d4d981be6532ae39f03327e13#diff-c70a88ce1c8376af239b0b2083d1d3c3) isn't matched before.

funkyfuture:
okay, i amended the requested changes and rebased onto #123.

nir0s:
I apologize, but there's been an oversight. Please fix the conflicts so that I can merge this.

funkyfuture:
rebased.

distro.py
```diff
@@ -925,7 +931,8 @@ def _lsb_release_info(self):
         elif rc == 127:  # Command not found
             return {}
         else:
-            if sys.version_info[0:2] >= (2, 7):
+            warn(err.decode('ascii'))
```
andy-maier: Why the warning when an exception is raised anyway?

distro.py
andy-maier: I think it would also work to make the release group in the regexp optional (on vacation with just the phone so I cannot test it). However, this elif path looks generally good to me. So I'm fine with this change.

funkyfuture: it still can be changed if it produces errors. i think simplicity beats purity here.

andy-maier: Agreed

funkyfuture:
i found the stderr useful at some point while debugging. i changed it back to be silent and adapted the signatures according to the Pythons. however i feel there could be an `LSBError` with all four attributes, otherwise testing would be more cumbersome.

nir0s:
We'll definitely handle it if it comes up.

Issue #133, opened 2016-07-23
Add the ability to run the module directly


nir0s:
For people who don't want to use Python directly, it might be a good idea to provide a way to run:

```shell
python -m distro
```

I think the output should be that of `distro.info()` but can't decide whether we should allow the `pretty` and `best` flags.

Maybe we should start from info and, according to user demand, add additional options.

nir0s:
#134 implements this and contains some last code adjustments for the 1.0.0 release.

MartijnBraam:
Maybe it's better to output something more human readable when running distro directly and some command line flag to output a json format if you want to parse it with bash. 

I've used the distro multiple times to find out what kind of Linux box I'm currently on.

If added a simple command line interface in 9c1143f3b6487c963fc1d89c7b675c3d7742913f that outputs the pretty name / version by default and outputs json when you supply the --json parameter

nir0s:
Good idea. Just thought we're start from simple. But I agree, this is better.

nir0s:
#134 solves this by adding a new CLI

PR #134, opened 2016-07-23
Some minor code adjustments plus a CLI

nir0s:


nir0s:
There's a problem with using argparse here. It's not supported by default in 2.6 (which we should still support due to the % of companies using it).

asottile:
For argparse you can do conditional requirements (such as [this example](https://github.com/pre-commit/pre-commit/blob/ea05189c28084e42014b20eae3a1130ea14c9d93/setup.py#L50))

andy-maier:
From looking at the code, it seems that test case `TestCli`assumes that the `distro`command outputs JSON format, but it does that only when invoked with the `-j` option.

andy-maier:
What do people think about an lsb_release compatible cli for distro?

nir0s:
In what sense?

nir0s:
As suggested by @asottile , do we agree that argparse should be conditional and that if someone who uses 2.6 doesn't have argparse it becomes a dependency?

MartijnBraam:
I think that's the better option, I personally find that argparse produces a way better command line tool than optparse.

The oldest python version I used is 2.7 and I don't have any experience with optional dependencies so I don't know the pitfalls but it seems like a good solution.

asottile:
Another option would be to drop 2.6 support (which has been end of lifed for several years).

nir0s:
Many companies still use 2.6. See the PR. I added `argparse` as a conditional requirement.
Also, added an entry point so that you can simply run `distro` instead of `python -m distro`.

nir0s:
If we're going to release 1.0.0, I think we need this in..

nir0s:
solves #133

PR #135, opened 2016-08-03
Use `bytes` invariantly.

asottile:
`bytes` was introduced in python2.6 to aid in porting to python3 by providing
a single name to refer to the "binary string" type.


andy-maier:
+1

PR #136, opened 2016-08-03
Use TOXENV environment variable

asottile:


andy-maier:
+1

nir0s:
Please fix conflicts

asottile:
Rebased to fix conflicts :)

nir0s:
This is somewhat less explicit. For instance. travis shows:

```
pip 8.1.2 from /home/travis/virtualenv/python3.5.2/lib/python3.5/site-packages (python 3.5)
```

Not saying that we shouldn't merge.. but It's somewhat harder to see the environment like that during tests.

asottile:
Previously you were setting python for the tox executable's sake only. I've
removed that bit since it's unnecessary.  The messaging you are seeing
comes from travis-ci directly.  I don't know of a way to turn it off

PR #150, opened 2016-11-15
Get UNIXCONFDIR from environment first, fallback to /etc fixes #149

### LINKED ISSUE #149: _UNIXCONFDIR is not configurable
When running distro confined in a Snap, `/etc` is not accessible for read access and instead is located at `$SNAP/etc`. Since `_UNIXCONFDIR` is hard coded it's impossible in modules, like pip, to address this relocation.

I'd like to suggest an environment variable, `UNIXCONFDIR` as a means to override this behavior, where the default remains the same.

marcoceppi:
fixes #149

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -41,7 +41,7 @@ import subprocess
 if not sys.platform.startswith('linux'):
     raise ImportError('Unsupported platform: {0}'.format(sys.platform))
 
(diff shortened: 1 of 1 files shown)


nir0s:
Thanks. Can you please add a test for that?

Issue #139, opened 2016-10-05
LICENSE file


xsuchy:
Can you please include LICENSE file in tar file, next time you will do the release? Probably by putting it in setup.py. It will easy its packaging in Fedora (I am doing it right now).

ignatenkobrain:
no, it should not be installed, it should be distributed along with PyPI tarball.

bachp:
I would also like to see this included as I'm packaging this for OpenEmbedded. Thanks.

nir0s:
v1.0.2 now contains the license file.

Issue #121, opened 2016-04-21
Use twine to release to PyPI


andy-maier:
I learned that the `twine` package has two advantages over `setup.py upload`:
- It uses encrypted communication with PyPI (setup.py added that too at some point, but I think not for Pyhton 2.6).
- It creates metadata on PyPI for the package dependencies. This results in a human readable section on the PyPI page, and I believe it also improves the machine-readable metadata about the dependencies.

nir0s:
I actually started using twine just recently. It's really nice. I'm all for it.

andy-maier:
I recently switched to twine in another project, and can contribute a PR to distro, soon.

nir0s:
FYI, I already released a wheel using twine.. just didn't change the Makefile to do that. You're invited :)

PR #154, opened 2017-01-12
Repo Organization

nir0s:


nir0s:
Resolves #139 - LICENSE file is now a part of the MANIFEST
Resolves #121 - Twine based publishing is now in the Makefile

Issue #157, opened 2017-01-14
Replace nose with pytest


nir0s:
Will be more comfortable to work with

PR #158, opened 2017-01-15
Replace nose with pytest

nir0s:


nir0s:
I'd be happy if someone could review this? It's very straightforward, but should still be looked at. If it takes too long, i'll just merge it.

nir0s:
Resolves #157

PR #163, opened 2017-01-27
Race condition in _parse_distro_release_file

### LINKED ISSUE #162: distro fails because it can't read /etc/image_version file
* Distro version: 1.0.2
* Python version: 3.4.3
* Operating system: Ubuntu 14.04

### Description:

On my system there is a file (out of my control) named /etc/image_version . This file is NOT world-readable. When using distro 1.0.2 in one of my virtualenvs (which I expect to be "isolated" from the system) distro will crash because it can't read said file.

I think the PermissionError should just be catched and ignored...

This issue was originally seen with pip (that vendors distro) and was reported here: https://github.com/pypa/pip/issues/4249

### What I've run:

```
testuser@host:~/tmp$ virtualenv -p python3 pip-test
Running virtualenv with interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in pip-test/bin/python3
Also creating executable in pip-test/bin/python
Installing setuptools, pip...done.
testuser@host:~/tmp$ source pip-test/bin/activate
(pip-test)testuser@host:~/tmp$ pip install distro
Downloading/unpacking distro
  Downloading distro-1.0.2-py2.py3-none-any.whl
Installing collected packages: distro
Successfully installed distro
Cleaning up...
(pip-test)testuser@host:~/tmp$ distro
Traceback (most recent call last):
  File "/home/testuser/tmp/pip-test/bin/distro", line 7, in <module>
    from distro import main
  File "/home/testuser/tmp/pip-test/lib/python3.4/site-packages/distro.py", line 1051, in <module>
    _distro = LinuxDistribution()
  File "/home/testuser/tmp/pip-test/lib/python3.4/site-packages/distro.py", line 596, in __init__
    self._distro_release_info = self._get_distro_release_info()
  File "/home/testuser/tmp/pip-test/lib/python3.4/site-packages/distro.py", line 996, in _get_distro_release_info
    distro_info = self._parse_distro_release_file(filepath)
  File "/home/testuser/tmp/pip-test/lib/python3.4/site-packages/distro.py", line 1016, in _parse_distro_release_file
    with open(filepath) as fp:
PermissionError: [Errno 13] Permission denied: '/etc/image_version'
```

SethMichaelLarson:
Also happens to fix #162 :)
Instead of checking if the file exists just try and open it and if we can't open it (For whatever reason, if it's not there, no permissions, etc) then we return as if it wasn't there.

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -1012,12 +1012,13 @@ class LinuxDistribution(object):
         Returns:
             A dictionary containing all information items.
         """
-        if os.path.isfile(filepath):
+        try:
             with open(filepath) as fp:
                 # Only parse the first line. For instance, on SLES there
                 # are multiple lines. We don't want them...
                 return self._parse_distro_release_content(fp.readline())
-        return {}
+        except (OSError, IOError):
+            return {}
 
     @staticmethod
     def _parse_distro_release_content(line):


nir0s:
LGTM. I want someone else (@andy-maier?) to look at it to verify that I'm not missing any potential breaking behavior here.

PR #167, opened 2017-02-09
Make distro more robust to unreadable files that end with "-release"

becker33:
We found sites that add other files that end with "-release" in /etc. When those files are world readable, distro properly recognizes them as irrelevant and continues. When they are not readable, distro throws an IOError due to missing file permissions.

This PR catches that exception and simply continues to the next candidate file to identify the release file.


distro.py
```diff
@@ -993,13 +993,17 @@ def _get_distro_release_info(self):
...
-                        return distro_info
+                    try:
+                        filepath = os.path.join(_UNIXCONFDIR, basename)
+                        distro_info = self._parse_distro_release_file(filepath)
+                        if 'name' in distro_info:
+                            # The name is always present if the pattern matches
+                            self.distro_release_file = filepath
+                            distro_info['id'] = match.group(1)
+                            return distro_info
+                    except IOError:
```
SethMichaelLarson: I would change this line to:
```python
except (OSError, IOError):
```
for compatibility reasons. :)

SethMichaelLarson:
Also should probably add a test for this?

nir0s:
@becker33, yup, please add a test.

SethMichaelLarson:
This is looking a lot better @becker33 :+1:

becker33:
I looked into why I couldn't reach the except clause. I had been working from a release prior to #163 being merged. #163 fixed issue #162, which is similar to my issue. The except clause is never used because #163 added error handling in _parse_distro_release_file().

I will close this as redundant.

Issue #152, opened 2016-12-07
Minimal information detected when lsb-release isn't installed


lenormf:
Hi,

I hit this bug while using the `distro` module in a Debian Stretch `docker` container:

```python
>>> distro._distroi.linux_distribution()
('Debian GNU/Linux', '', '')
>>> distro._distroi.info()
{'like': '', 'codename': '', 'id': 'debian', 'version': '', 'version_parts': {'build_number': '', 'major': '', 'minor': ''}}
```

Expected output:

```python
>>> distro.linux_distribution()
('Debian GNU/Linux', 'testing', 'stretch')
>>> distro.info()
{'like': '', 'version': 'testing', 'codename': 'stretch', 'id': 'debian', 'version_parts': {'major': '', 'build_number': '', 'minor': ''}}
```

The /etc/os-release:
```shell
PRETTY_NAME="Debian GNU/Linux stretch/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

Everything worked fine when I installed the `lsb-release` package, but I was under the impression that I wouldn't need it in the first place.

HTH.

nir0s:
I assume you mean that you would expect to see "stretch/sid" in the codename?

lenormf:
Sorry, I hadn't included the expected output. I think the `codename` should indeed be `stretch/sid`, or just `stretch` as seems to be the case on some systems.

nir0s:
Unfortunately, this seems to be something that will have to be addressed as part of #109.

In debian 8, the codename appeared in parentheses and in stretch, it isn't. That's inconsistent with almost all distributions. It's also rather sad that they're not providing the version anymore.


```
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support/"
BUG_REPORT_URL="https://bugs.debian.org/"
```

I will start working on a distro inconsistencies framework soon, at which point i'll address this as well.

asottile:
fwiw, the latest `stretch` docker image has an os-release file that looks like this:

```
root@a83fb62a5bc1:/# cat !$
cat /etc/os-release
[...]

Issue #78, opened 2016-03-16
Testcase for Linux Mint is missing


andy-maier:
Linux Mint (ID: `linuxmint`) is one of the distros with a reliable ID. A testcase should be added.

If you want to help, it is sufficient to post to this issue:
* The content of the `/etc/os-release` file, if any.
* The file names and content of the `/etc/*release` and `/etc/*version` files, if any.
* The output of the command: `lsb_release -a`, if available.

Napsty:
/etc/os-release:

	$ cat /etc/os-release 
	NAME="Ubuntu"
	VERSION="14.04.3 LTS, Trusty Tahr"
	ID=ubuntu
	ID_LIKE=debian
	PRETTY_NAME="Ubuntu 14.04.3 LTS"
	VERSION_ID="14.04"
	HOME_URL="http://www.ubuntu.com/"
	SUPPORT_URL="http://help.ubuntu.com/"
	BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"

/etc/lsb-release:

    $ cat /etc/lsb-release
    DISTRIB_ID=LinuxMint
    DISTRIB_RELEASE=17.3
    DISTRIB_CODENAME=rosa
    DISTRIB_DESCRIPTION="Linux Mint 17.3 Rosa"

/etc/upstream-release/lsb-release:

    $ cat /etc/upstream-release/lsb-release 
    DISTRIB_ID=Ubuntu
    DISTRIB_RELEASE=14.04
    DISTRIB_CODENAME=trusty
    DISTRIB_DESCRIPTION="Ubuntu 14.04 LTS"

/etc/debian_version:

    $ cat /etc/debian_version
    jessie/sid

Output of lsb_release -a:

    $ lsb_release -a
    No LSB modules are available.
    Distributor ID:	LinuxMint
    Description:	Linux Mint 17.3 Rosa
    Release:	17.3
    Codename:	rosa

andy-maier:
Claudi,
I started integrating this and only now realized that the data in the os-release file is not about LinuxMint, but about its underlying Ubuntu.

The approach we have in the `ld` package is to have a precedence between the different data sources: os-release file -> lsb_release command -> distro release file.

As a result, the consolidated accessor functions such as `ld.id()` or `ld.name()` will return the Ubuntu data.

`ld` also provides single source accessor functions to access a particular data source specifically, for example `ld.lsb_release_info()` will return the LinuxMint data provided by the lsb_release command.

[...]

Issue #177, opened 2017-04-01
Support for Windows and Mac OS.


SethMichaelLarson:
Brought this up in Gitter and apparently this has had work done on it in a separate branch. (Link that here?). I've had multiple times where I wanted to have the amount of information about a Windows or Mac OS platform as the distro module gives but `platform.mac_ver` and `platform.win32_ver` are pretty deficient in. This thread is mostly for discussing how this would be done. cc: @nir0s

nir0s:
So, My idea was that we implement a class per OS type. Distro will identify the OS it's running on (Darwin, nt, linux, etc..) and generate a `distro` object for that OS type when it is imported.

When you run `distro.id()`, for example, it would map what would be the ID in Windows to the relevant information provided by the OS.

This will aid users is making OS dependent decisions without having to know which OS they're running on.

nir0s:
That being said, this means that to keep a consistent API as well have to come up with a good mapping from each OS type to the names of the functions already in distro.

SethMichaelLarson:
I would suggest changing `linux_distribution` to just `distribution`? The rest map pretty well actually. All Windows and Mac distributions have a codename, id='windows', 'osx', version, name, pretty name.

nir0s:
`distro.linux_distribution` is meant to replace `platform.linux_distribution`.. so I think it's kinda weird to remove it now, though there's no problem with deprecating it at some point.

If we want `distribution`, I'd add it on top of `linux_distribution` while `linux_distribution` only returns for linux and `distribution` returns for any.

SethMichaelLarson:
I'd say that something like that would work fine. Do we only define that method when on Linux (forcing `distro` users to check with `hasattr`) or have `linux_distribution` throw an error on non-Linux (forcing `distro` users to catch or check the system type first) or map `linux_distribution` to `distribution` just for backwards-compatibilities sake?

nir0s:
Um, it seems more natural to me to return empty for `linux_distribution` if you're not on linux and maybe throw a warning. Mapping might create some ambiguity since you're getting information on a non-linux system in what you'd expect to work only on linux. The reason because of which I like the return empty thing, is because distro tends to return (much like `platform`'s implementation) empty results when it can't find something. This behavior would correspond with the natural flow distro's takes instead of unexpectedly erroring out only in this specific use case.

SethMichaelLarson:
Oh sure! That can definitely be done. So just a `Distribution` object that has all empty values.

[...]

PR #178, opened 2017-04-01
Fixing how distro deals with inconsistent *-release files.

SethMichaelLarson:
So this PR is a minimal (and IMO rather hacky) solution to the two cases where distro has trouble with getting the right information from *-release files. The two related issues are #78 for Linux Mint and #152 for Debian stretch/sid when lsb-release isn't installed. I don't think this solution should be merged, might be a good place for some discussion about the proper direction.

If I were to be able to make more changes it might be nice to have `distro.LinuxDistribution()` do all it's parsing, assigns values, and then return a class that does minimal logic and spits out those values when they're queried.  Could still keep same interface including `os_release_attr`, `lsb_release_attr`, etc. Would allow for a cleaner layout of where special cases happen rather than have each special case occur within the `id()`, `codename()` and `version()` functions. I think this would be a much better route to go, especially for when future inconsistencies are found.

Thoughts?

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -602,6 +602,7 @@ class LinuxDistribution(object):
         self._lsb_release_info = self._get_lsb_release_info() \
             if include_lsb else {}
         self._distro_release_info = self._get_distro_release_info()
+
         self._resolve_distro_inconsistencies()
 
     def __repr__(self):
@@ -682,14 +683,21 @@ class LinuxDistribution(object):
         For details, see :func:`distro.version`.
         """
         versions = [
-            self.os_release_attr('version_id'),
+            self._preferred_release_attr({'os-release': 'version_id',
+                                          'lsb-release': 'release',
+                                          'distro-release': 'version_id'})[0],
             self.lsb_release_attr('release'),
             self.distro_release_attr('version_id'),
             self._parse_distro_release_content(
-                self.os_release_attr('pretty_name')).get('version_id', ''),
-            self._parse_distro_release_content(
                 self.lsb_release_attr('description')).get('version_id', '')
         ]
+
+        # Linux Mint has Ubuntu's version in os-release so exclude it from this list.
+        if self.id() != 'linuxmint':
+            versions.insert(1, self.os_release_attr('version_id'))
+            versions.insert(4, self._parse_distro_release_content(
+                self.os_release_attr('pretty_name')).get('version_id', ''),)
+
         version = ''
         if best:
             # This algorithm uses the last version in priority order that has
@@ -762,10 +770,9 @@ class LinuxDistribution(object):
 
         For details, see :func:`distro.codename`.
         """
-        return self.os_release_attr('codename') \
-            or self.lsb_release_attr('codename') \
-            or self.distro_release_attr('codename') \
-            or ''
+        return self._preferred_release_attr({'os-release': 'codename',
+                                             'lsb-release': 'codename',
+                                             'distro-release': 'codename'})[0]
 
     def info(self, pretty=False, best=False):
         """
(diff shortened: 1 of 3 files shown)


nir0s:
This is cool and certainly will allow us to solve some inconsistencies but will also somewhat pollute the code as we find more and more inconsistent implementations because it changes the default implementation in the base class.

My idea is to write a class per inconsistent distribution which will inherit `LinuxDistribution` and implement only the inconsistent behavior while using the default class's parsing methods. 

So, for instance, in Linux Mint's use case, the method under `LinuxDistrubtion` which returns the ordering of the files, would return something else. There will be a single method under `LinuxDistribution` which would call the relevant class's implementation based on the different inconsistencies which would replace the base class's attribute with the new implementation.

Any main method in the base classe (`id()`, `version()`, etc...) will call that method and if there is an inconsistency, that attribute will be replaced and returned. If there is no inconsistency, it will return the original value.

Should I provide an example, or is that clear?

nir0s:
An example:

replace:

`_distro = LinuxDistribution()`

with:

```python
class Centos7(LinuxDistribution):
    def codename(self):
        return 'Ridiculous'

    def version(self, pretty=False, best=False):
        return '7.9'


def _get_implementation():
    main = LinuxDistribution()
    if main.id() == 'centos' and main.version() == '7':
        return Centos7()

    return LinuxDistribution()


_distro = _get_implementation()
```

I think it's a very clean solution and wonderfully easy for developers to support.

WDYT?

SethMichaelLarson:
I agree completely on the solution. This PR was more to be exploratory and to spark discussion. :) My only concern is that now developers can't just call `LinuxDistribution` to get the proper object to describe their system like they could in the past.

nir0s:
Why not?

SethMichaelLarson:
If they do on a system with inconsistencies they won't receive `Centos7Distribution`.

nir0s:
Oh you mean if their distribution has inconsistencies. True.

SethMichaelLarson:
So I currently see two ways of dealing with it, one more transparent but also breaks functionality in the case that anyone is subclassing the `LinuxDistribution` type. I don't know why anyone would, but they could do it in theory.

We could rename `LinuxDistribution` to `_LinuxDistribution` and then make a function `LinuxDistribution` that does exactly what `_get_implementation()` does. This would break things if anyone is subclassing but also would make it so no code besides that needs to change.

We could make `_get_implementation` a public method? Maybe rename it to `get_distribution()`? But then all code that calls `LinuxDistribution` directly needs to be rewritten to use the new helper.

As crazy as it sounds I think I'm more in favor of option number 1, even though it would probably necessitate a major version bump...

SethMichaelLarson:
And if you think about it option 1 lends itself well to if and when we decide to implement Windows and Mac OS functionality. Have a single function `Distribution()` that determines what object you actually get.

nir0s:
While I agree that it would be "safer" to fork the class to a function, it would also be misleading and non-future-proof. While distro isn't that much widely used, I'd rather release a major version breaking the previous one and explicitly stating that it breaks stuff in BIG RED :)

Using `_get_implementation`, I think we've being very explicit. I think that making it a public method which returns an object (not necessarily `LinuxDisribution` obviously), would be safer.

We will say something like:

By default, you can still use `distro.id()`, etc..
If you want to get the relevant object for your distribution, simply do:

`dist = distro.get_implementation()` instead of `dist = distro.LinuxDistrubtion()`.

Yes, it breaks backwards, but I'd release distro 2.0 for that to be so clean.

SethMichaelLarson:
The name is fine with me! Sounds good. :)

nir0s:
I think so. Makes sense.. But maybe we should get some more feedback about win and osx support?

SethMichaelLarson:
@nir0s I'd love some feedback. :) Do we have an opinion from any upstream like pip?

nir0s:
@dstufft? @xavfernandez? @andy-maier? Your input?

SethMichaelLarson:
Just an FYI to reviewers first reading this PR: This PR isn't what needs review. Just need feedback on adding Windows/OSX and the proposed method of mitigating Linux distro inconsistencies by @nir0s :) Thanks!

SethMichaelLarson:
Since we're talking about it here as well as in another issue I'll tag it: #177

xavfernandez:
Concerning pip, the only place we are using `distro` is here: https://github.com/pypa/pip/blob/3cdd162/pip/download.py#L94-L107 , to construct a user-agent with accurate information about the user's system.
So concerning Windows/OSX, it might be interesting to have more information than we currently provide but it's not clear what it could be.
Concerning the idea to `write a class per inconsistent distribution which will inherit LinuxDistribution` it seems quite clean 👍

nir0s:
I guess we can start working on the proposed solution and keep discussing additional OS related stuff in the dedicated issue.

SethMichaelLarson:
That sounds good to me, I'll start a little work on a v2 branch and we can take it from there.

PR #181, opened 2017-04-03
Add MandrivaLinux test case

SethMichaelLarson:
So I'm not sure if we should have Mandriva Linux have the ID of 'mandriva' or 'mandrivalinux'. Inside of lsb-release it lists MandrivaLinux as the ID.


tests/test_distro.py
```diff
@@ -1377,8 +1376,29 @@ def test_openelec6_release(self):
             'minor_version': '0',
         }
         self._test_outcome(desired_outcome)
-        
-        
+
+    def test_mandriva2011_release(self):
+        desired_outcome = {
+            'id': 'mandrivalinux',
```
nir0s: It's so bad that different files return different information. We really need that framework.. meh :(

SethMichaelLarson:
So it looks like I accidentally added a copy of `mandriva2011` folder that I was using to gather all the files to the actual git commit. I'll open a PR to remove that folder.

Issue #184, opened 2017-04-18
CloudLinux <7 appears with ID = 'rhel'


SethMichaelLarson:
Just opening this issue as we now know from #180 that CloudLinux will appear with an id of `rhel` in releases before CloudLinux 7. Putting this issue here to track when we release v2.

PR #180, opened 2017-04-02
Add test cases for CloudLinux 5, 6, and 7

SethMichaelLarson:
I've got test cases here for CloudLinux 5, 6, and 7. CloudLinux 5 and 6 only have `redhat-release` (making them show up with and id of `rhel`) and 7 has `redhat-release` and `os-release` which makes the id show up correctly as `cloudlinux`. None of the three install `lsb-release` by default. So this is another case of inconsistent ids which will have to be solved in a different PR.

This PR solves issue #75.

PR #185, opened 2017-04-18
Work for v2 (Windows+Mac OS support, Linux inconsistencies)

SethMichaelLarson:
Been doing a little work for v2 in my own branch. If anything else needs to be added I'll add it to the checklist. Here's what I have so far:

### Checklist
- [x] distro is now a package instead of one file. **This means that Python 2.6 can't run it as a module** via `python -m distro` or `python -m distro -j` but since we're incrementing a major version and this feature isn't used by our most important upstream pip this is probably fine?
- [x] Split testing up into "CLI", "Linux", "Windows".
- [x] Changed `LinuxDistribution` object to inherit from a new object `Distribution` which is the base for all distribution info objects.
- [x] Added a new function called `get_distribution()` which gets the `Distribution` object for the current OS platform. This is where sub-classes of `LinuxDistribution` are created if there is an inconsistency between one of the *-release sources (such as `LinuxMintDistribution`.)
- [x] `ImportError` is raised if not on Windows, Linux, or Mac OS platforms.
- [x] `MacOSDistribution` needs to be completely finished and needs testing. I don't have my own Mac OS to get real values for this but I'm sure most of the values can be guessed after seeing a few real values of `platform.mac_ver()`.
- [x] `CloudLinuxDistribution` needs to be created to deal with inconsistencies for Cloud Linux once #180 has been merged.
- [x] `DebianDistribution` for when debian doesn't use parens for it's codename in "stretch/sid".
- [ ] Go through all comments, docs, and package info and remove all mentions distro of being Linux-only.
- [ ] Resolve all other inconsistencies that I've missed and update tests to reflect.
- [x] Create an `EmptyDistribution` object to return when `distro.linux_distribution()` is called on a non-Linux platform.
- [ ] Enable Mac OS builders on Travis CI (?)
- [ ] Get tests to cover the module 100% and then start enforcing 100% coverage. (?)
- [ ] Add all changes to the CHANGELOG and explain the new major release.

### Questions to Resolve

- [ ] Should calling `distro.linux_distribution()` on a non-Linux platform emit a warning?
- [ ] Should we add `distro.windows_distribution()` and `distro.mac_distribution`? I'm in favor of saying "No" to this to encourage users to use `distro.distribution()` and only keeping `distro.linux_distribution()` for legacy reasons.
- [ ] Should we raise an `ImportError` if we're on a system we don't recognize or should we just return an `EmptyDistribution` object for unknown platforms?
- [ ] Continue to support 1.0.x versions? Cut a v1 branch on GitHub and work from that with master as v2 after this is approved?

### References

#177 #178 #180 #184

cc: @nir0s


nir0s:
This is a wonderful effort and I like the implementation. It's very clean. Thank you.

Several comments.

1. I think we should keep it a single module instead of a package. I think that distro should be as portable as possible so that it can even be copied to servers on the fly to retrieve distribution information. I'm certainly going to be using fabric to deploy distro to servers to retrieve info and making distro a package makes it very hard if not impossible. I don't see why the entire codebase couldn't fit in a single file. It still has a very very small footprint.
2. I think we have to separate the inconsistency framework and the windows/mac/ support into multiple requests. These are very different things and we need multiple reviewers to make sure everything fits. We also need the ability to merge only some of the code. I'd start from a framework PR for inconsistencies on top of which we could add multi-OS support.
3. We should still verify that it makes sense to break backward compat with regard to the `LinuxDistribution` class. What I proposed was just an idea :)

@xavfernandez @andy-maier @MartijnBraam @funkyfuture. Would love your input here.

SethMichaelLarson:
Sure thing, I'll break this whole effort into two requests. :) Thanks for the review!

distro/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+The ``distro`` package (``distro`` stands for Linux Distribution) provides
+information about the Linux distribution it runs on, such as a reliable
```
funkyfuture: that'd nee to be adjusted.

distro/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+    * ``like``:  The result of :func:`distro.like`.
+
+    * ``codename``:  The result of :func:`distro.codename`.
+
+    For a description of the *pretty* and *best* parameters, see the
+    :func:`distro.version` method.
+    """
+    return _distro.info(pretty, best)
+
+if platform.system() == 'Linux':
```
funkyfuture: imports should be on the top of the module and be absolute. older Python versions have `from __future__ import absolute_imports` available.

distro/__init__.py
```diff
@@ -0,0 +1,425 @@
...
+    from ._linux import get_distribution
+elif platform.system() == 'Windows':
+    from ._windows import get_distribution
+elif platform.system() == 'Mac OS':
+    from ._mac_os import get_distribution
+else:
+    from ._distribution import EmptyDistribution
```
funkyfuture: rather `UnknownDistribution` or `Unidentified…`?

SethMichaelLarson:
@funkyfuture I've split this into another PR (#185) for just Linux. I'm keeping this PR here mostly for the check boxes. Thank you for your review! :)

SethMichaelLarson:
I'd still like to get these changes merged. Is this still on the table?

nir0s:
@SethMichaelLarson, I apologize for the extremely lengthy delay (personal issues..). I'd like to get back to this and merge it. For that, it would be great to get the single module going. WDYT?

dsully:
I'd also like to see this - particularly macOS.

SethMichaelLarson:
My pull requests for this project have dwelled too long. Anyone is welcome to finish the work out if they'd like. I left the branch un-deleted.

Issue #109, opened 2016-04-19
Provide way to deal with distro inconsistencies


andy-maier:
There are Linux distros that have ionconsistent information in the data sources that  `ld` uses.

For example , see issue #78 about Linux Mint. In that case, the precedence implemented by `ld` causes Linux Mint to be shown as an Ubuntu. While it is fairly compatible to Ubuntu, it is not Ubuntu. It has its own name, id, version (Linux Mint 17.3 is based upon Ubuntu 14.04), and code name ("rose" for Linux Mint 17).

I think that instead of asking the distros to become consistent, that calls for some mechanism in `ld` to deal with that.

We already parse all three data sources if available, and we could find out that the (normalized) IDs are different, and if so, trigger some special-case processing. In the case of Linux Mint, the special processing would be to ignore those data sources that advertise Ubuntu.

I guess the trick is to implement this without getting into endless if-elif-else statements.

nir0s:
I guess we could easily implement a class per special case. Thing is, if we don't do if else here to explicitly provide the distros we know are "problematic", we'll be forcing checking for that consistency throughout all distros even if 99% of them don't have that problem. I'm not saying we shouldn't, just raising it as an issue.

My intuition tells me that a "bug which happens once is not a bug". We can specifically specify Linux Mint's condition in the code and return different values for it. If we stumble across additional distros containing this issue, we can start discussing a mechanism.

The implementation doesn't have to be ugly. We could state a list of distros which override the type of information returned and put mint in that list.

sethmlarson:
To make things easier for us and more consistent across multiple tools imo we should treat bugs in /etc/*-release files as bugs upstream instead of something actionable in Distro. Thoughts from other maintainers?

brejoc:
> To make things easier for us and more consistent across multiple tools imo we should treat bugs in /etc/*-release files as bugs upstream instead of something actionable in Distro. Thoughts from other maintainers?

While I would absolutely sympathize with this approach, this might mean that things stay broken for a long time. It's sometimes hard to motivate distros to change something in their *-release files, and even if they are willing to, the release cycles might pretty long.

nir0s:
With @brejoc here. The reason why you need distro in the first place is to abstract long-term ugliness.

sethmlarson:
I'm not sure that this fix is needed though, wouldn't we see more chatter / +1s on this issue if there were widespread problems with distributions identifying incorrectly? Quoting @nir0s from 2016:

[...]

PR #186, opened 2017-04-19
Distro Inconsistencies Framework

sethmlarson:
These are some of the changes for working towards v2.

### Changes

- Added `get_implementation()` which will choose which `LinuxDistribution` sub-class to use based on known inconsistencies within Linux distros.
- Added `LinuxMintDistribution` which will not use most values from `/etc/os-release` if the file is not changed from upstream Ubuntu.
- Added `CloudLinuxDistribution` which will always return an id of `cloudlinux` even if there are only `/etc/redhat-release` files available.
- Added `DebianDistribution` which will detect the new format that Debian is using for codenames within `/etc/os-release` `PRETTY_NAME` value in the case that `lsb_release` is not available. (Note I still don't have the complete set of `/etc/*-release` files from Debian 9, I didn't actually check. I was just going off the data that was provided in #109.

### References

#109 #180 #184 

cc: @nir0s @xavfernandez @andy-maier @MartijnBraam @funkyfuture

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -842,13 +842,13 @@ class LinuxDistribution(object):
         Returns:
             A dictionary containing all information items.
         """
-        if os.path.isfile(self.os_release_file):
+        try:
             with open(self.os_release_file) as release_file:
                 return self._parse_os_release_content(release_file)
-        return {}
+        except (OSError, IOError):
+            return {}
 
-    @staticmethod
-    def _parse_os_release_content(lines):
+    def _parse_os_release_content(self, lines):
         """
         Parse the lines of an os-release file.
 
@@ -1097,6 +1097,19 @@ class CloudLinuxDistribution(LinuxDistribution):
         return 'cloudlinux'
 
 
+class DebianDistribution(LinuxDistribution):
+    """ Starting with Debian 9 (stretch) the information inside of /etc/os-release
+    does not follow the same standards for codenames that all other distros follow
+    thus custom detection of Debian's codename is required. """
+    def _parse_os_release_content(self, lines):
+        props = super(DebianDistribution, self)._parse_os_release_content(lines)
+        if 'codename' not in props and props.get('pretty_name', ''):
+            match = re.search(r'\s([^\s\d]+)$', props['pretty_name'])
+            if match:
+                props['codename'] = match.group(1)
+        return props
+
+
 def get_implementation(*args):
     ld = LinuxDistribution(*args)
     os_release_id = _normalize_id(ld.os_release_attr('id'), NORMALIZED_OS_ID)
@@ -1104,7 +1117,9 @@ def get_implementation(*args):
     if os_release_id == 'ubuntu' and lsb_release_id == 'linuxmint':
         return LinuxMintDistribution(*args)
     elif ld.id() == 'rhel' and 'CloudLinux' in ld.name():
-        return CloudLinuxDistribution()
+        return CloudLinuxDistribution(*args)
+    elif ld.id() == 'debian' and ld.codename() == '':
+        return DebianDistribution(*args)
     else:
         return ld
 
--- a/distro.py
+++ b/distro.py
@@ -898,7 +898,7 @@ class LinuxDistribution(object):
                         codename = codename.strip('()')
                         codename = codename.strip(',')
                         codename = codename.strip()
-                        # codename appears within paranthese.
+                        # codename appears within parenthesis.
                         props['codename'] = codename
                     else:
                         props['codename'] = ''
@@ -1092,20 +1092,19 @@ class LinuxMintDistribution(LinuxDistribution):
 
 
 class CloudLinuxDistribution(LinuxDistribution):
-    """ CloudLinux <7 doesn't provide an lsb-release or os-release file,
-    only a redhat-release file so it's id is incorrectly detected as `rhel`.
-    """
     def id(self):
+        """ CloudLinux <7 doesn't provide an lsb-release or
+        os-release file, only a redhat-release file so it's id
+        is incorrectly detected as `rhel`. """
         return 'cloudlinux'
 
 
-class DebianDistribution(LinuxDistribution):
-    """ Starting with Debian 9 (stretch) the information inside of
-    /etc/os-release does not follow the same standards for codenames
-    that all other distros follow thus custom detection of Debians
-    codename is required. """
+class DebianStretchDistribution(LinuxDistribution):
     def _parse_os_release_content(self, lines):
-        props = (super(DebianDistribution, self).
+        """ Debian 9 (stretch) does not have the codename
+        in parenthesis or with a comma unlike all other
+        distributions. This stub """
+        props = (super(DebianStretchDistribution, self).
                  _parse_os_release_content(lines))
         if 'codename' not in props and props.get('pretty_name', ''):
             pretty_name = props['pretty_name']
@@ -1120,7 +1119,7 @@ class DebianDistribution(LinuxDistribution):
         return props
 
 
-def get_implementation(*args):
+def get_distribution(*args):
     """ Gets the proper implementation of LinuxDistribution after attempting
     to detect distro inconsistencies. Passes all args through to the resulting
     LinuxDistribution object. """
@@ -1133,13 +1132,13 @@ def get_implementation(*args):
         return LinuxMintDistribution(*args)
     elif ld.id() == 'rhel' and 'CloudLinux' in ld.name():
         return CloudLinuxDistribution(*args)
-    elif ld.id() == 'debian':
-        return DebianDistribution(*args)
+    elif ld.id() == 'debian' and ld.codename() == '':
+        return DebianStretchDistribution(*args)
     else:
         return ld
 
 
-_distro = get_implementation()
+_distro = get_distribution()
 
 
 def main():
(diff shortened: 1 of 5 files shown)


distro.py
```diff
@@ -1074,7 +1074,72 @@ def _parse_distro_release_content(line):
...
+            # a codename of 'GNU/Linux'.
+            pretty_name = pretty_name.replace(props.get('name', ''), '')
+
+            match = re.search(r'\s([^\s\d]+)$', pretty_name)
+            if match:
+                props['codename'] = match.group(1)
+        return props
+
+
+def get_implementation(*args):
```
funkyfuture: is this supposed to be a public interface? the name doesn't tell me what this function is about, and i get the docstring's meaning also rather vaguely.
this is the first time i encounter the term 'implementation' in the context of distributions. as a non-native speaker.

SethMichaelLarson: It's public-ish. It's probably not going to be used by a single non-power user because most users will just use `distro.id()`, `distro.name()`, etc. The word implementation was chosen because @nir0s suggested it in a prior thread. I thought that naming it `get_distribution` would clash too closely to the other function that I know will be defined `distribution()` when we add in support for Windows and Mac OS.

SethMichaelLarson:
Sorry, I added you as a reviewer here because you were pinged for comments on the previous PR.

distro.py
```diff
@@ -1074,7 +1074,72 @@ def _parse_distro_release_content(line):
...
+class CloudLinuxDistribution(LinuxDistribution):
+    """ CloudLinux <7 doesn't provide an lsb-release or os-release file,
+    only a redhat-release file so it's id is incorrectly detected as `rhel`.
+    """
+    def id(self):
+        return 'cloudlinux'
+
+
+class DebianDistribution(LinuxDistribution):
+    """ Starting with Debian 9 (stretch) the information inside of
```
nir0s: In general, I think the docstring should be inside each method of a class so that we can explain things even if we have multiple inconsistencies per distribution.

distro.py
nir0s: @funkyfuture, what do you suggest?

distro.py
```diff
@@ -1074,7 +1074,72 @@ def _parse_distro_release_content(line):
...
+
+class CloudLinuxDistribution(LinuxDistribution):
+    """ CloudLinux <7 doesn't provide an lsb-release or os-release file,
+    only a redhat-release file so it's id is incorrectly detected as `rhel`.
+    """
+    def id(self):
+        return 'cloudlinux'
+
+
+class DebianDistribution(LinuxDistribution):
```
SethMichaelLarson: I was wondering your opinion on this as well. I will convert it to be version-specific. It doesn't return a version because I used to a very constrained set of /etc/*-release files in order to get this issue to show itself. Specifically not using lsb-release as was the case in the GitHub issue where this was reported.

distro.py
SethMichaelLarson: Sure thing!

distro.py
funkyfuture: looking at the docs and the code, i would say this function does `get_linux_distribution` or `detect_…`?!

nir0s: Not necessarily Linux considering that we're thinking of adding windows and mac, but yes, `get_distribution` would be fine. I'm convinced.

distro.py
nir0s: Right. I'll open an issue for the version thing.

nir0s: Regarding the version specific thing.. it would be great if a version-agnostic class of each distribution where there are constant inconsistencies would exist for version specific ones to inherit from.. so if you found that there are distros where we need a version agnostic implementation, I wouldn't add a version to those implementations.

SethMichaelLarson: Okay, I'll add a `DebianStretchDistribution` class. :) I'm guessing because Linux Mint has the os-release problem it'll stay with a single class until Linux Mint fixes that issue?

SethMichaelLarson:
Okay I'm working on the comments now! :steam_locomotive:

distro.py
SethMichaelLarson: I'm grabbing the latest Debian 9 ISO and getting all the release and version files, then the `TestOverall` testcase for Debian 9 will return the version.

SethMichaelLarson:
This should be ready for re-review @nir0s :)

distro.py
```diff
@@ -1074,7 +1074,71 @@ def _parse_distro_release_content(line):
...
+    return table.get(distro_id, distro_id)
+
+
+class LinuxMintDistribution(LinuxDistribution):
+    """ LinuxMint will have the same /etc/os-release file
+    as it's upstream Ubuntu so we should not use any info
+    from that file except `ID_LIKE`. """
+    def os_release_attr(self, attribute):
+        if attribute == 'id_like':
+            v = super(LinuxMintDistribution, self).os_release_attr(attribute)
```
funkyfuture: no need for the symbol `v` here.

distro.py
```diff
@@ -1074,7 +1074,71 @@ def _parse_distro_release_content(line):
...
+            return v
+        return ''
+
+
+class CloudLinuxDistribution(LinuxDistribution):
+    def id(self):
```
funkyfuture: if `LinuxDistribution.id` was a wrapped as `property` (which makes sense due to its static-ness during runtime), `id` could simply be a class-property then:

```python
class CloudLinuxDistribution(LinuxDistribution):
    id = 'cloudlinux'
```

funkyfuture:
concerning the `id` (or any other runtime-determined properties), i recently came to a pattern to make such lazy-evaluated that would in this case look like this:

```python
class LinuxDistribution:
    @property
    def id(self):
        result = do_heuristics()
        setattr(self, 'id', result)
        return result
```

i haven't yet faced a situation or doubt why that would be a bad idea. but i haven't seen it in the wild yet.

distro.py
SethMichaelLarson: Line length constraints. :(

distro.py
SethMichaelLarson: Are we changing `id` from a function to a property? The same could be said about all other `name()`, `version()`, etc

distro.py
funkyfuture: you can use a line-break in the return statement.

SethMichaelLarson: I'm aware, I just thought this looked better than this
```python
return (super(LinuxMintDistribution, self)
        .os_release_attr(attribute))
```

or this

```python
return super(LinuxMintDistribution,
             self).os_release_attr(attribute)
```

funkyfuture: rather 

```python
return super(LinuxMintDistribution, self)\
    .os_release_attr(attribute)
```

i actually don't know whether the current state would be optimized upon bytecode-compilation. if so, it doesn't matter anyway.

SethMichaelLarson: This would create a linting error because there is a hanging indent that has a multiple of 4 spaces that's not a code indent. 

This type of thing isn't optimized away in byte-code, two extra codes for `STORE FAST` and `LOAD FAST` before the return.

distro.py
SethMichaelLarson: I'm in favor of keeping `id` a function, it would keep things homogenous between the module level `id()` and the `Distribution.id()`.

funkyfuture: sure, it should be canonical. yet, these are all constants.

SethMichaelLarson: They happen to be constants for `CloudLinuxDistribution` but that's not necessarily the case for all, most distributions have to get the information about `id` from `os_release_attr`, `lsb_release_attr`, and `distro_release_attr` function calls.

I know we can just use the `@property` decorator but there are more "constant" values than just `id`.

funkyfuture: that's my point (not concerning a particular class, but in general) - all these values are constants during runtime, especially from the client code's perspective.

SethMichaelLarson: So you're proposing caching the value? There's nothing wrong with a function returning something that is actually a constant. See the class of functions this library is aiming to replace: `platform.linux_distribution()`, `platform.version()` etc. Those values won't change during runtime but they're all still functions.

funkyfuture: i mean both: changing the design and as a side-effect to 'cache'.

```python

def _get_id():
    …

ID = _get_id()

__all__ = ['ID']
```

though i'd argue that no uppercase reads cleaner in the client code: `f'Running on {distro.name}.'`

@nir0s shall i open an issue regarding the design discussion? though i can't promise to layout a detailed proposal soon, i see the opportunity to introduce breaking changes with a new major release. sorry for bloating this up here.

SethMichaelLarson: This is probably going to be a little trickier to test compared to our current model of doing things.

SethMichaelLarson:
Would still like these changes to be merged. Ping?

nir0s:
I deeply apologize. I'll get back to this asap. @SethMichaelLarson 

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

SethMichaelLarson:
No worries! :)

nir0s:
@SethMichaelLarson, excuse me for the extremely lengthy delay (personal issues..). I'd like to get this merged while addressing in the docs that there are breaking changes. Would you mind solving the conflicts?

Thank you!

SethMichaelLarson:
On vacation currently, when I return I'll take a look at this again.

zztalker:
Please, explain. Do you decide to do not implement that?

nir0s:
@sethmlarson @zztalker, I'd actually like to get this in, but I don't currently have the time. If someone is willing to take it and address the issues, I'll merge and release ASAP.

Issue #190, opened 2017-08-02
Add slackbuilds package link


nir0s:
https://slackbuilds.org/repository/14.2/python/python-distro/

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

Issue #188, opened 2017-06-09
Add Centos distribution link 


nir0s:
https://pkgs.org/download/python2-distro

_Sent from my LGE Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #192, opened 2017-10-11
Add CentOS and Slackbuilds package links

nir0s:


nir0s:
Fixes #188, Fixes #190

PR #195, opened 2017-11-28
Remove Official Python26 support

nir0s:
* Remove pyt2.6 in all testing environments.
* Remove implicit argparse installation for py2.6.


nir0s:
Note, Python3.3 tests fail over codecov installation. Checked it manually and everything is fine. Must be cache related since I installed their 3.3 version manually and installed codecov successfully.

PR #197, opened 2017-12-03
Remove shell overhead

asottile:


asottile:
python3.3 is failing, this addresses that: https://github.com/nir0s/distro/pull/199

nir0s:
Brilliant.

PR #200, opened 2017-12-04
Lazy load LinuxDistribution

### LINKED ISSUE #198: Considerable import-time side-effects
Importing distro performs (at the least) a subprocess call, this can be considerably slow and if distro isn't directly used could impact application startup time.

Most of this stems from the assignment of `distro._distro`:

https://github.com/nir0s/distro/blob/ff1699d63b3ba93411d339fa1b19836684c6f420/distro.py#L1077

This _could_ be avoided either by lazily instantiating the `_distro` object, or lazily assigning the attributes of `LinuxDistribution`.

I've attached a profile, here's some high level numbers though:

### status quo
```
$ time python -c 'import distro'

real	0m0.297s
user	0m0.112s
sys	0m0.024s
```

### removing module-time side-effects (naive)

```diff
--- a/distro.py
+++ b/distro.py
@@ -1074,7 +1074,7 @@ class LinuxDistribution(object):
         return distro_info
 
 
-_distro = LinuxDistribution()
+#_distro = LinuxDistribution()
 
 
 def main():
```

```
$ time python -c 'import distro'

real	0m0.019s
user	0m0.016s
sys	0m0.000s
```

### profile

Obtained using:
```
echo 'import distro' > test.py
python -m cProfile -o prof.pstats test.py
gprof2dot --skew=.1 -wn3 prof.pstats | dot -Tsvg -o prof.svg
```

Using [yelp's fork of gprof2dot (`pip install yelp-gprof2dot`)](https://github.com/Yelp/gprof2dot)

nir0s:
Fixes #198. 

* `LinuxDistribution` is no longer loaded on import.
* `distro.init()` will return a `LinuxDistribution` object and instantiate it idempotently.
* All info functions will lazily call `init()`.

@asottile 

```
$ time python -c "import distro"

real	0m0.051s
user	0m0.044s
sys	0m0.004s
```

Need to actually write a test for `init()` and verify lazy loading and idempotence, but I don't have time for that right now.

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -94,6 +94,14 @@ _DISTRO_RELEASE_IGNORE_BASENAMES = (
     'system-release'
 )
 
+_DISTRO = None
+
+
+def init():
+    global _DISTRO
+    _DISTRO = _DISTRO or LinuxDistribution()
+    return _DISTRO
+
 
 def linux_distribution(full_distribution_name=True):
     """
@@ -120,7 +128,7 @@ def linux_distribution(full_distribution_name=True):
     method normalizes the distro ID string to a reliable machine-readable value
     for a number of popular Linux distributions.
     """
-    return _distro.linux_distribution(full_distribution_name)
+    return init().linux_distribution(full_distribution_name)
 
 
 def id():
@@ -195,7 +203,7 @@ def id():
       command, with ID values that differ from what was previously determined
       from the distro release file name.
     """
-    return _distro.id()
+    return init().id()
 
 
 def name(pretty=False):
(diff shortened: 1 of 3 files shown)


asottile:
I was actually thinking this could be done just with lazy properties on the `LinuxDistribution` object -- let me throw together a quick patch and see how you like it :)

asottile:
here's the alternative implementation: https://github.com/nir0s/distro/pull/201

nir0s:
#201 is a better solution. Closing.

PR #201, opened 2017-12-04
Lazily load the LinuxDistribution data

### LINKED ISSUE #198: Considerable import-time side-effects

asottile:
Resolves #198

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -511,6 +511,21 @@ def distro_release_attr(attribute):
     return _distro.distro_release_attr(attribute)
 
 
+class cached_property(object):
+    """A version of @property which caches the value.  On access, it calls the
+    underlying function and sets the value in `__dict__` so future accesses
+    will not re-call the property.
+    """
+    def __init__(self, f):
+        self._fname = f.__name__
+        self._f = f
+
+    def __get__(self, obj, owner):
+        assert obj is not None, 'call {} on an instance'.format(self._fname)
+        ret = obj.__dict__[self._fname] = self._f(obj)
+        return ret
+
+
 class LinuxDistribution(object):
     """
     Provides information about a Linux distribution.
@@ -576,6 +591,9 @@ class LinuxDistribution(object):
           `distro release file`_ that is actually used as a data source. The
           empty string if no distro release file is used as a data source.
 
+        * ``include_lsb`` (bool): The result of the ``include_lsb`` parameter.
+          This controls whether the lsb information will be loaded.
+
         Raises:
 
         * :py:exc:`IOError`: Some I/O issue with an os-release file or distro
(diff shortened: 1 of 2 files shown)


nir0s:
Definitely better solution than mine.

distro.py
```diff
@@ -511,6 +511,17 @@ def distro_release_attr(attribute):
     return _distro.distro_release_attr(attribute)
 
 
+class cached_property(object):
```
nir0s: I think we should add docstrings here to explain what we're doing.

distro.py
```diff
@@ -591,26 +602,20 @@ def __init__(self,
         self.os_release_file = os_release_file or \
             os.path.join(_UNIXCONFDIR, _OS_RELEASE_BASENAME)
         self.distro_release_file = distro_release_file or ''  # updated later
-        self._os_release_info = self._get_os_release_info()
-        self._lsb_release_info = self._get_lsb_release_info() \
-            if include_lsb else {}
-        self._distro_release_info = self._get_distro_release_info()
+        self.include_lsb = include_lsb
```
nir0s: as it's a part of the API, it should be added to the documented.

distro.py
asottile: done!

distro.py
asottile: done!

PR #202, opened 2017-12-04
Don't raise ImportError on non-linux platforms

asottile:
In order to be able to use `distro` in cross-platform programs without duplicated import guards, `distro` shouldn't raise on `import`.

For what it's worth, after this patch `distro` acts _mostly_ in the same way as `platform.linux_distribution` on non-linux platforms:

```
>>> platform.linux_distribution()
('', '', '')
>>> distro.id(), distro.version(), distro.codename()
('', '', '')
```

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -38,9 +38,6 @@ import argparse
 import subprocess
 
 
-if not sys.platform.startswith('linux'):
-    raise ImportError('Unsupported platform: {0}'.format(sys.platform))
-
 _UNIXCONFDIR = os.environ.get('UNIXCONFDIR', '/etc')
 _OS_RELEASE_BASENAME = 'os-release'
 


asottile:
This will fix my patch here on non-linux platforms: https://github.com/asottile/pip-custom-platform/pull/15

nir0s:
The reason this was added is partly because `platform.linux_distribution` returns empty when it doesn't know how to identify the distro it's running on. In a Windows/Darwin case, it's not that we can't identify it, it's just that it's not supposed to.

asottile:
I think that's fine, at the very least it's the status quo for the `platform` module.

This patch makes using `distro` in a cross platform way much easier and definitely makes migrating from old `platform`-based code dead simple.

nir0s:
@asottile btw, it would be awesome to get your feedback on `https://github.com/cloudify-cosmo/wagon`.

asottile:
@nir0s looks pretty neat!  Seems to be trying to solve the same problems as `pip-custom-platform` too.

I'd be curious to see the full workflow of how `wagon` gets used -- for `pip-custom-platform` it's largely:
- build wheels for all supported platform combinations for all supported packages
- place those wheels on an internal pypi server
- `s/pip/pip-custom-platform/g` in installation scripts

PR #203, opened 2017-12-22
Stdout of shell should be decoded with sys.getfilesystemencoding()

doloopwhile:
Hello.

Errors are raised when `$LANG` is not utf-8 (ex. `LANG=ja_JP.EUCJP`).

This cause a bug in `pip`.
https://github.com/pypa/pip/issues/4917

Would you review it ?

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -920,7 +920,7 @@ class LinuxDistribution(object):
                 stdout = subprocess.check_output(cmd, stderr=devnull)
             except OSError:  # Command not found
                 return {}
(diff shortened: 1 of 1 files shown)


SethMichaelLarson:
Hey @doloopwhile! Thanks for making this contribution, it looks like a positive change for me.
Currently it looks like our Travis build is failing for unrelated reasons, I'll see if I can fix this. :tada:

SethMichaelLarson:
Also going to ping @nir0s as this issue should probably mean a new release of `distro` so that `pip` can vendor it.

PR #230, opened 2018-12-12
Prefer the VERSION_CODENAME field of os-release to parsing it from VERSION

abadger:
Preferring the [codename field] lets distributions accurately encode their
codename preferences instead of us having to parse them from the VERSION
in ways that may be error prone.

This came up in two known instances:
* Ubuntu 16.04 has a VERSION string of "16.04.1 LTS (Xenial Xerces)"
  and a VERSION_CODENAME (Actually UBUNTU_CODENAME because this was before Ubuntu adopted the standard) of "xenial".  distro was mistakenly returning
  "xenial xerces" because of the way the VERSION string was formatted.
* [Fedora 21+] do not have a code name.  Our parser was returning the
  variant because Fedora sticks the variant in parenthesis in the
  VERSION field.  Fedora <strike>30+</strike>28+ will add an [empty VERSION_CODENAME] field so
  that we can recognize they do not have a code name.

[codename field]: https://www.freedesktop.org/software/systemd/man/os-release.html#VERSION_CODENAME=
[Fedora 21+]: https://lists.fedoraproject.org/pipermail/advisory-board/2013-October/012209.html
[empty VERSION_CODENAME]: https://src.fedoraproject.org/rpms/fedora-release/pull-request/58#


sgallagher:
> Fedora 30+ will add an [empty VERSION_CODENAME](https://src.fedoraproject.org/rpms/fedora-release/pull-request/58#) field so that we can recognize they do not have a code name.

Just for the record, we also merged this for F28 and F29, so anyone with a fully up-to-date system will have the correct empty VERSION_CODENAME as well.

abadger:
Updated the commit message to reflect that this PR will correct the codename on Fedora 28 and greater.

nir0s:
This is wonderful. Thanks @abadger. Thing is, I'd rather have the `ubuntu_codename` specific handling a part of https://github.com/nir0s/distro/pull/186. I'm trying to avoid handling distribution specific inconsistencies within the main code.

nir0s:
Hmm. Having given it some more thought, it seems like merging #186 will take a while. I'll merge this, and once the inconsistencies framework is in, we'll move this.

abadger:
Cool.  Makes sense to me.

PR #231, opened 2018-12-13
Rebase python2.6 support

abadger:
This PR rebases the python2.6 support branch to current master.  It's the first step towards actually making a common python-2.6 compatible distro codebase.


abadger:
@nir0s @SethMichaelLarson  This is in reference to: https://github.com/nir0s/distro/pull/206  I'll be opening another PR on top of this to actually add python-2.6 support.

I've just done the work to bundle distro into ansible for our 2.8 release: https://github.com/ansible/ansible/pull/35985 and it would make sense to me to have a common place for 2.6 compatibility to land.  I can't promise to be a great deal of help on this but I'll do what's needed to get the initial version of python-2.6 support passing in Travis and take a look at rebasing onto newer versions whenever a new Ansible release is in the pipeline.  If other people who are bundling join in, that would make me happy as it would be less work for me.

PR #233, opened 2018-12-19
Return _uname_info from the uname_info() method

### LINKED ISSUE #229: Missing return statement
```python
In [1]: from distro import uname_info                                              

In [2]: print(uname_info())                                                        
None
```
What does the function below ?
https://github.com/nir0s/distro/blob/d20463809ce8ebb24177ea6df5e371be6d6963e1/distro.py#L868

abadger:
Fixes #229

I'm sorry, I don't understand how the test cases work well enough to see why the existing uname_info tests failed to pick up this problem otherwise I would have adapted them to test this case too.  I'm guessing that they test the _uname_info() method directly instead of testing the uname_info() method but I don't quite see where that would be happening.

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -872,6 +872,7 @@ class LinuxDistribution(object):
 
         For details, see :func:`distro.uname_info`.
         """
+        return self._uname_info
 
     def os_release_attr(self, attribute):
         """


nir0s:
Great catch. The tests are actually testing uname attributes directly, which is wrong. I'll merge this, but we should definitely attend to the missing tests.

Issue #161, opened 2017-01-15
Add support for detecting a distro from a rootfs even on another OS


pombredanne:
I would be interested to run a distro detection based on some rootfs-like archive for aLinux distro and that on any OS. It could be a a plain pre-imaged rootfs, a container image, a VM image or similar and I would like to detect what it is regardless of which OS `distro` runs on.

pombredanne:
So essentially I would modify distro to accept an optional path to the root of a distro FS defaulting to `/` so the behavior would be the same as today by default and would not attempt to run `lsb_release` unless the path is the default '/' and the current OS is Linux.

pombredanne:
... and make a few paths relative to that provide root path rather than absolute.
Do you have any appetite for including this type of behavior?

nir0s:
Well, you're the first only asking for such behavior, so I can't say that I've had any appetite for it just yet. It went through my mind that it would be nice to support remote distro identification in the sense that you could just provide a machine to identify a distribution on.

pombredanne:
perfect. I will consider a PR .

hartwork:
`lsb_release` does not seem to support a custom root directory. So for the callto `lsb_release -a` (https://github.com/nir0s/distro/blob/1a7d01b095f757754b511591ebe428bd57e19d34/distro.py#L917) it seems to me that root permissions and using `chroot` would be needed. Unless the call to `lsb_release` can be avoided. I'm curious what you come up with.

pombredanne:
@hartwork What is `lsb_release`?  it depends! 

This is a spec AFAIK. Tool-wise we have many implementations.

At https://sourceforge.net/projects/lsb and now at http://bzr.linuxfoundation.org/loggerhead/lsb/3.1/si/annotate/head:/lsb_release/src/lsb_release this is a shell script.

In Debian and Ubuntu this is a Python script:
https://anonscm.debian.org/cgit/collab-maint/lsb.git/tree/lsb_release.py
http://packages.ubuntu.com/yakkety/lsb-release


In Fedora and Red Hat, the lsb shell script is used as a base:
http://pkgs.fedoraproject.org/repo/extras/redhat-lsb/redhat-lsb-4.1-1.tar.bz2/9c1e474cb81c0ee1941c3129e0e50ea1/

AFAIK, other distros each use their own minor variations or do not have it at all.


So the strategy could be:
1. skip running `lsb_release` entirely.
[...]

PR #247, opened 2019-11-12
Detect distro from arbitrary rootfs root_dir #161

pombredanne:
This introduce a new optional root_dir argument to contruct a
LinuxDistribution object. When provided, this is used as if it were
the root of the filesystem when looking up for files.
This also adds a --root-dir option to the command line.
This is a fix for #161

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


hartwork:
The command line interface driven by `main` in `distro.py` does not support a custom root directory yet. Might be just the right time to turn module scope `_distro` into something that is obtained on demand so that the host system is not inspected unconditionally when importing the module. Maybe wrap every access in some singleton-like internal getter?

pombredanne:
@hartwork re

> The command line interface driven by main in distro.py does not seem to gain support for a custom root directory from this pull request yet. 

Would you like an option?

> Might be just the right time to turn module scope _distro into something that is obtained on demand so that the host system is not inspected unconditionally when importing the module. Maybe wrap every access in some singleton-like internal getter?

I could not agree more. Having a global created on import here does not make sense to me https://github.com/nir0s/distro/blob/cdfe85d15bd366820db6a1cfdc6cf9a0a5de7e37/distro.py#L1189

A lot of the module api could be made simpler IMHO.

hartwork:
> Would you like an option?

That would be plain awesome, yes :smile:

pombredanne:
@hartwork there you go the latest push has a cli option now

pombredanne:
@hartwork NB: I added your "reported-by" there https://github.com/nir0s/distro/pull/247/commits/ab9d3c1fa7fcc71056700841b6a273b18af20b6d ... I assume this is OK with you!

tests/test_distro.py
```diff
@@ -85,6 +89,33 @@ def test_cli_json(self):
...
+            "like": "",
+            "version": "30",
+            "version_parts": {
+                "build_number": "",
+                "major": "30",
+                "minor": ""
+            }
+        }
+        '''
+        desired_output = json.loads(desired_output)
```
hartwork: I noticed we're parsing hard-coded JSON here while then comparing high-level objects. How about removing the JSON indirection layer by going straight to Python nested dictionaries:
```
desired_output = {
    'codename': '',
    'id': 'fedora',
    'like': '',
    'version': '30',
    'version_parts': {
        'build_number': '',
        'major': '30',
        'minor': '',
    },
}
```
As a side-effect, we can now re-introduce trailing commas for better diffs if those lines are touched in the future. What do you think?

tests/test_distro.py
```diff
@@ -432,6 +463,22 @@ def test_cloudlinux7_os_release(self):
...
+    """Test that a LinuxDistribution can be created using an arbitrary root_dir
+    on all OSes.
+    """
+
+    def setup_method(self, test_method):
+        dist = test_method.__name__.split('_')[1]
+        root_dir = os.path.join(DISTROS_DIR, dist)
+        self.distro = distro.LinuxDistribution(
+            os_release_file='',
+            distro_release_file='non',
```
hartwork: Why does it say `'non'` here?

tests/test_distro.py
pombredanne: Sure, I was a tad lazy for this. !

tests/test_distro.py
pombredanne: That's something I carried over from the test class I subclass here https://github.com/nir0s/distro/blob/ab9d3c1fa7fcc71056700841b6a273b18af20b6d/tests/test_distro.py#L151

I scratched my head as this does not make sense. I think this is meant to be a `non`-existing release file, but I did not dig why you would want that. For the sake of simplicity I kept that as otherwise the tests are not behaving the same way as if you were passing no `distro_release_file` at all. ... which is really weird as this should not matter. @nir0s any clue?

hartwork: I dug a bit in the code as well just now: There is special handling for `distro_release_file` being "true" — `if self.distro_release_file:` — and the method `_parse_distro_release_file` swallows all file access errors in silence. So `non` seems more intentional to me now but I'd use a filename like `no_such_file` or so to better communicate that intention.

pombredanne:
@nir0s ping? Do you mind to review this? Thanks!

nir0s:
I've been away. I'll find time to review this soon.

Thanks for this!

pombredanne:
ping ... this is now stale.. I will rebase

pombredanne:
@nir0s I updated to the latest master.

pombredanne:
@nir0s If you can tell if you may or may not merge that, it would be useful. If not I will have to maintain my own fork and push it as distro2 to PyPI.

nir0s:
I see no reason not to get this in. I'll give it a look during the weekend.

pombredanne:
@nir0s Thank you ++! :bow:

PR #262, opened 2020-04-21
Honors `/usr/lib/os-release` when `/etc/os-release` is not available

HorlogeSkynet:
Hey over here :wave:

So, there is ~~a bug~~ an unexpected behavior with **distro** in very "lite" environments, when for example even `/etc/os-release` is not available.

> One-liner to reproduce : `docker run archlinux:latest /bin/bash -c 'pacman --noconfirm -Sy python-distro && python3 -m distro'`

It actually appears that we _should_ be dealing with `/usr/lib/os-release` in such cases ([reference](https://www.freedesktop.org/software/systemd/man/os-release.html#Description)).

This patch implements the expected behavior, and adds a very specific test case to verify it.

Bye :bow:


HorlogeSkynet:
Note : the patch _seems_ back-portable to the `python2.6-support` branch.

HorlogeSkynet:
Rebased against new `master`, fixing conflicts to include changes brought by #247.

(cc @nir0s)

HorlogeSkynet:
Up @nir0s please :roll_eyes:

HorlogeSkynet:
Up... again @nir0s (cc @hartwork maybe ?) 😕

HorlogeSkynet:
One more "up" for 2020.

hartwork:
> Up... again @nir0s (cc @hartwork maybe ?) confused

I don't have capacity for proper review here atm, I'll leave this to @nir0s for now.

nir0s:
Fellas, I truly apologize, but I indeed hardly have any capacity to deal with issues nowadays.

I'd appreciate adding collaborators. Any volunteers?

HorlogeSkynet:
Well, thanks for your answer. I can't go full-time on it of course, but as I now maintain a [project completely based on yours](https://github.com/HorlogeSkynet/archey4/blob/2b956faf069cad5fe8f230325bfc8d70429540c2/setup.py#L28), I'm indirectly interesting in `distro`'s "health condition".
So I'm okay for sharing work with other(s), if it's really needed.

Bye, take care.

hartwork:
> Fellas, I truly apologize, but I indeed hardly have any capacity to deal with issues nowadays.

I think it's good and fair to be realistic and open about it, thanks for that. And thanks for your work on distro in general :pray: At the same time I think it's a bit sad also e.g. because of the [9000+ projects dependening on `distro` today](https://github.com/nir0s/distro/network/dependents) (on GitHub alone) including important players like SaltStack, Let's Encrypt Certbot, docker-compose and package manager conan. So the 154 stars up here are only the tip of the iceberg.

If I may ask, are your capacity constraints temporary or expected long-term? If long-term, maybe it makes sense to turn this project into a new GitHub org then. It's rather easy to do technically — create the org, then do a repo transfer (which moves all issues and PRs just like that, then add people) — and it may communicate better that the project has become the effort of a new team if you're saying goodbye long-term. At least one other trusted person will also need PyPI access for a full transition. I hope I'm being respectful here, I didn't intend otherwise. What do you think?

> I'd appreciate adding collaborators. Any volunteers?

Speaking for myself, I will not be able to shoulder much load here but if there are other active contributors, I may be able to join and help a bit myself too (after my current sprint on https://github.com/git-big-picture/git-big-picture has cooled down more).

README.md
```diff
@@ -89,10 +89,10 @@ The `distro` package implements a robust and inclusive way of retrieving the
 information about a distribution based on new standards and old methods,
 namely from these data sources (from high to low precedence):
 
-* The os-release file `/etc/os-release`, if present.
+* The os-release file `/etc/os-release` if present, with a fall-back on `/usr/lib/os-release` if needed.
 * The output of the `lsb_release` command, if available.
 * The distro release file (`/etc/*(-|_)(release|version)`), if present.
-* The `uname` command for BSD based distrubtions.
+* The `uname` command for BSD based distributions.
```
hartwork, line 95: With my picky hat on: I'm all for fixing typos but I'd vote to do that in a dedicated commit so that people after you and me will be able to look at pure semantic changes without skip-reading over spelling fixing. Just my 2 cents.

distro.py
```diff
@@ -37,8 +37,8 @@
 import argparse
 import subprocess
 
-
 _UNIXCONFDIR = os.environ.get('UNIXCONFDIR', '/etc')
+_UNIXLIBSDIR = os.environ.get('UNIXLIBSDIR', '/usr/lib')
```
hartwork, line 41: I find the `S` in `UNIXLIBSDIR` weird. I would vote for `UNIXLIBDIR` without `S` or `_UNIXUSRLIBDIR`. Just an idea.

distro.py
```diff
@@ -654,9 +654,10 @@ def __init__(self,
         self.root_dir = root_dir
         self.etc_dir = os.path.join(root_dir, 'etc') \
             if root_dir else _UNIXCONFDIR
-        self.os_release_file = os_release_file or \
-            os.path.join(self.etc_dir, _OS_RELEASE_BASENAME)
-        self.distro_release_file = distro_release_file or ''  # updated later
+        self.usr_lib_dir = os.path.join(root_dir, 'usr/lib') \
+            if root_dir else _UNIXLIBSDIR
+        self.os_release_file = os_release_file
```
hartwork, line 659: I worry about the change in semantics of `self.os_release_file` a bit. Previously, it was guaranteed to be the path to the file about to be tried to read. Now it can still be the empty string :thinking: 
I'm aware that constructors/initializers should not do much and that maybe doing the `os.path.isfile` detection loop (that I have seen further down) in here *could* not be ideal, I'm not sure. What do you think?

distro.py
```diff
@@ -672,8 +673,7 @@ def __repr__(self):
             "_os_release_info={self._os_release_info!r}, " \
             "_lsb_release_info={self._lsb_release_info!r}, " \
             "_distro_release_info={self._distro_release_info!r}, " \
-            "_uname_info={self._uname_info!r})".format(
-                self=self)
+            "_uname_info={self._uname_info!r})".format(self=self)
```
hartwork, line 676: With my picky hat on: If this change is needed for a green CI, I vote to move to a dedicated commit. If it's not needed for a green CI, I vote for reverting. Thanks!

distro.py
```diff
@@ -935,9 +935,23 @@ def _os_release_info(self):
...
+            if os.path.isfile(self.os_release_file):
+                return _open_and_return_info(self.os_release_file)
+        else:
+            # If not, try known "regular" locations of os-release files.
+            for os_release_file_candidate in (
+                    os.path.join(self.etc_dir, _OS_RELEASE_BASENAME),
+                    os.path.join(self.usr_lib_dir, _OS_RELEASE_BASENAME)):
+                if os.path.isfile(os_release_file_candidate):
+                    return _open_and_return_info(os_release_file_candidate)
+
```
hartwork, line 954: While the current code checks for `os.path.isfile`, with regard to [the spec](https://www.freedesktop.org/software/systemd/man/os-release.html) it seems to me that we should respect a symlink with an existing target — no? Then `os.path.exists` might be the function of choice. What do you think?

PS: I think the DRY problem can be addressed in a simpler way. What about:

```python
for os_release_file in (
        self.os_release_file,
        os.path.join(self.etc_dir, _OS_RELEASE_BASENAME),
        os.path.join(self.usr_lib_dir, _OS_RELEASE_BASENAME),
        ):
    if os_release_file and os.path.isfile(os_release_file):
        with open(os_release_file) as f:
            return self._parse_os_release_content(f)

return {}
```

docs/index.rst
```diff
@@ -184,7 +184,8 @@ Os-release file
 ===============
 
 The os-release file is looked up using the path name ``/etc/os-release``. Its
-optional additional location ``/usr/lib/os-release`` is ignored.
+optional additional location ``/usr/lib/os-release`` would be looked up for if
+the former couldn't be read.
```
hartwork, line 188: I see one more place in `index.rst` that _may_ need an extension like that. It's in section `Compatibility` and it reads:

> * The content of the `/etc/os-release` file, if any.

tests/test_distro.py
```diff
@@ -122,23 +123,26 @@ class DistroTestCase(object):
     represented in the `DISTROS` subtree.
     """
 
-    def setup_method(self, test_method):
+    def setup_method(self, _):
```
hartwork, line 126: With my picky hat on: No objections but my vote for moving to commit dedicated to unused variables.

tests/test_distro.py
```diff
@@ -687,7 +691,6 @@ def test_empty_release(self):
     def test_unknowndistro_release(self):
         self._setup_for_distro(os.path.join(TESTDISTROS, 'distro',
                                             'unknowndistro'))
-
```
hartwork, line 690: With my picky hat on: No objections but my vote for moving to a commit dedicated to whitespace changes.

hartwork (commented):
I had a closer look now. Given [what the spec says about `/usr/lib/os-release`](https://www.freedesktop.org/software/systemd/man/os-release.html) I think *merging some version of this pull request will be important* to do justice to the specification. So thanks to @HorlogeSkynet for working on this matter! :pray: 
Below are my findings. Some of them are picky… but with good intentions. I'm happy to learn I misunderstood or overlooked something. Best, Sebastian

distro.py
HorlogeSkynet: I hadn't opted for this solution as by instantiating a `LinuxDistribution` object with a speficied `os_release_file` attribute, the developer _might_ expect the loading of their file **only**.
So the DRY method imposed itself, as the logic _has to_ (?) be separated in two distinct branches.

README.md
HorlogeSkynet: I understand your point.
As I didn't think, back in April, to _take over_ (?) this project, I wanted to propose many improvements as possible. If I do, I'll propose them again in a dedicated PR, with dedicated commits 👌

distro.py
HorlogeSkynet: Second suggestion applied.

distro.py
HorlogeSkynet: Reverted.

tests/test_distro.py
HorlogeSkynet: Reverted.

tests/test_distro.py
HorlogeSkynet: Reverted.

distro.py
hartwork: I see how about:
```python
if self.os_release_file:
    candidates = [self.os_release_file]
else:
    candidates = [
        os.path.join(self.etc_dir, _OS_RELEASE_BASENAME),
        os.path.join(self.usr_lib_dir, _OS_RELEASE_BASENAME),
    ]

for os_release_file in candidates:
    if os_release_file and os.path.isfile(os_release_file):
        with open(os_release_file) as f:
            return self._parse_os_release_content(f)

return {}
```

distro.py
hartwork: I'm not sure about internal: It seems public API to me.

Before:
```console
# python -c 'import distro; print(distro.LinuxDistribution().os_release_file)'
/etc/os-release
```

After:
```console
# python -c 'import distro; print(distro.LinuxDistribution().os_release_file)'
<just a newline>
```

hartwork: What about something like this:

```python
if os_release_file:
    self.os_release_file = os_release_file
else:
    etc_dir_os_release_file = os.path.join(self.etc_dir, _OS_RELEASE_BASENAME) 
    usr_lib_os_release_file = os.path.join(self.usr_lib_dir, _OS_RELEASE_BASENAME)

    # NOTE: The idea is to respect order
    #       *and* have it set at all times for API backwards compatibility
    if os.path.isfile(etc_dir_os_release_file) or not os.path.isfile(usr_lib_os_release_file):
        self.os_release_file = etc_dir_os_release_file
    else:
        self.os_release_file = usr_lib_os_release_file
```

HorlogeSkynet: Well, I haven't thought about this very attribute as being part of the API. IMHO, it should be private 🤷

hartwork: I'm not the one to decide. Without leading underscores I cannot say it's not public without get in conflict with myself. Any ideas how to proceed?

HorlogeSkynet: Of course, it's currently _public_, but I guess it _should not_.
Two quick advanced GitHub search requests ([here](https://github.com/search?p=6&q=os_release_file+distro+language%3APython+NOT+filename%3Autil.py+NOT+filename%3Ametadata.py+NOT+filename%3Adistro.py+NOT+filename%3Autils.py&type=Code) & [there](https://github.com/search?q=distro.os_release_file+NOT+filename%3Autils.py&type=Code)) _seems_ to invalidate the fact that this attribute is _publicly_ being used (at least across GH's repositories).

Although, your idea to keep it up-to-date according to our internal logic is interesting too, we _only_ have to decide whether it's relevant or not.

Cheers 👋

hartwork: @nir0s we could benefit from your input here. Do you have a minute?

hartwork: @HorlogeSkynet I think in general our options are:

- (a) Treat `distro.LinuxDistribution.os_release_file` as current public API and change the PR to be more backwards compatible in semantics
- (b) Treat `distro.LinuxDistribution.os_release_file` as current public API by mistake, hide it away with leading underscores, increase the major version of distro to 2 to communicate the API break the semver way, and probably make a ton of other accidentally public API private while at it so that we don't break the API soon again.
- (c) Assume that `distro.LinuxDistribution.os_release_file` is of little interest to any user, ignore that it is current public API technically, change its current semantics, and see if we get (arguably justified) bug reports.

I have trouble standing behind (c) myself and (b) will probably not be fun or feasible, so I personally consider (a) the least worst.  Are there any other options?  What do you think?

HorlogeSkynet: I think you're 100% right and I would not have come to these abstract leads by myself.
Of course `(a)` would be less _breaking_ than the others.
If I were a "proper" maintainer here, I'd have (cleaned up the code and) proposed an implementation for `(b)` I guess.
About `(c)`, maybe there are "others" here familiar with the code base/the project that would be able to argue in favor of it ?

Anyway, feel free to edit my branch as required, I don't mind.
Bye 👋

hartwork: I don't have editing rights outside my own pull requests here. If you'd be okay to go with (a) would making `__init__` do https://github.com/nir0s/distro/pull/262#discussion_r552161650 and then making `_os_release_info` do plain `return _open_and_return_info(self.os_release_file)` would be an option to you? That would be (a) in practice in my mind, or at least one way to do it. If we had state, I could add an approving review and then maybe it gets merged this month with some luck.  What do you think?

HorlogeSkynet: Hey @hartwork, please find a new commit implementing `(a)` (and respecting `flake8` code-styles !). Bye, keep me posted :wave:

distro.py
```diff
@@ -938,6 +958,7 @@ def _os_release_info(self):
         if os.path.isfile(self.os_release_file):
             with open(self.os_release_file) as release_file:
                 return self._parse_os_release_content(release_file)
+
```
hartwork, line 961: With my picky hat on: If this (leftover) addition of a blank line could be reverted, that would be perfect.

hartwork (approved):
@HorlogeSkynet looks good to me now, thanks for your contribution and your patience! :pray:

HorlogeSkynet:
> @HorlogeSkynet looks good to me now, thanks for your contribution and your patience! 🙏

You're very welcome, sorry for this little leftover that didn't catch my attention sooner.
Bye 🙇

hartwork:
@nir0s this PR is ready and waiting for your review by now. Do you have a minute?

hartwork:
@nir0s any news?

hartwork:
@nir0s how can we move forward?

HorlogeSkynet:
So sad only pull requests "brandishing" the `fork` totem are able to get things move forward... 🙁

hartwork:
> @nir0s this PR is ready and waiting for your approval by now. Do you have a minute?

@nir0s please please

tests/resources/testdistros/distro/usrlibosreleaseonly/usr/lib/os-release
```diff
@@ -0,0 +1,3 @@
+NAME="usr_lib_os-release_only"
```
sethmlarson, line 1: Can we have a real-life `/usr/lib/os-release` example if one is available?

README.md
```diff
@@ -89,7 +89,7 @@ The `distro` package implements a robust and inclusive way of retrieving the
 information about a distribution based on new standards and old methods,
 namely from these data sources (from high to low precedence):
 
-* The os-release file `/etc/os-release`, if present.
+* The os-release file `/etc/os-release` if present, with a fall-back on `/usr/lib/os-release` if needed.
```
sethmlarson, line 92: I wonder if it makes sense to link to https://www.freedesktop.org/software/systemd/man/os-release.html here or somewhere?

distro.py
```diff
@@ -654,8 +655,27 @@ def __init__(self,
...
+            etc_dir_os_release_file = os.path.join(
+                self.etc_dir, _OS_RELEASE_BASENAME
+            )
+            usr_lib_os_release_file = os.path.join(
+                self.usr_lib_dir, _OS_RELEASE_BASENAME
+            )
+
+            # NOTE: The idea is to respect order **and** have it set
+            #       at all times for API backwards compatibility.
+            if os.path.isfile(etc_dir_os_release_file) \
```
sethmlarson, line 673: If there is no `/etc/os-release` and no `/usr/lib/os-release` let's set `self.os_release_file` back to the `/etc/os-release` path for backwards compatibility?

sethmlarson (changes requested):
Overall this looks like a good improvement. Had some comments for you:

distro.py
hartwork: I checked the if-guard again and what you're asking for is already happening. Am I missing something?

README.md
hartwork: Sounds good, but seems unrelated to this pull request.

tests/resources/testdistros/distro/usrlibosreleaseonly/usr/lib/os-release
hartwork: It would actually make the in-test data explain itself less. My vote against that, the test data seems perfectly fine to me.

README.md
sethmlarson: Not sure how you think it's unrelated, this was the reference given in the PR body?

distro.py
sethmlarson: You're right! Misread the conditions.

sethmlarson (dismissed):
Going to approve since the reference comment is optional.

README.md
HorlogeSkynet: IMHO the documentation needs a global update. Maybe we could address this in another PR ?

nir0s:
There are conflicts now, though. @HorlogeSkynet.

HorlogeSkynet:
> There are conflicts now, though. @HorlogeSkynet.

🙄 I'll try to rebase soon.

hartwork:
> roll_eyes I'll try to rebase soon.

@HorlogeSkynet let me know if you need help

HorlogeSkynet:
@nir0s I went for a manual merge, all good now ✅

hartwork (approved):

nir0s (approved):

PR #266, opened 2020-09-20
Travis-CI: Add Python 3.9-dev testing, migrate nightly to 3.10a

ferdnyc:
This PR adds a `python: 3.9-dev` test target, and migrates the `python: nightly` target to expect Python 3.10 (`TOXENV=py310`). This causes `nightly` to genuinely pass, where it was previously "passing" only due to being listed in `allow_failures`.

I left `nightly` in `allow_failures` all the same, since it will always be in flux to some extent.

I added `3.9-dev` to `allow_failures` as well, because at some point presumably 3.9 will be released. I figured that was the point where it makes sense to change `python: 3.9-dev` to `python: 3.9` and remove it from the `allow_failures` list.


.travis.yml
```diff
@@ -40,8 +40,12 @@ matrix:
       env: TOXENV=py38
       stage: test
 
+    - python: 3.9-dev
```
jdufresne, line 43: Since this PR was opened, 3.9 has been released. So we can use that now.

```suggestion
    - python: 3.9
```

.travis.yml
```diff
@@ -53,6 +57,7 @@ matrix:
       stage: test
 
   allow_failures:
+    - python: 3.9-dev
```
jdufresne, line 60:
```suggestion
```

jdufresne (commented):
The trove classifier could be added to `setup.py`:

        'Programming Language :: Python :: 3.9',

Python 3.9 could also be added to `tox.ini`.

ferdnyc:
> The trove classifier could be added to `setup.py`:
> 
> ```
>     'Programming Language :: Python :: 3.9',
> ```
> 
> Python 3.9 could also be added to `tox.ini`.

Done and done. While I was in there I dropped the now EOL'd Pythons 3.4 and 3.5 as well.

I left Python 2 / Python 2.7 (and tox's `py27`) alone for now. My inclination would be to drop those as well, but I'll leave it to others with more skin in that game.

ferdnyc:
LMK if you need me to squash the branch. It's already up to 4 commits, for about that many total lines of change. :rofl:

jdufresne (approved):

sethmlarson:
Closing this as it's no longer needed for Travis. However it'd definitely be worth opening to add 3.9 and 3.10-dev for GitHub Actions

PR #267, opened 2020-10-18
Add pretty_version and best_version to linux_distribution()

### LINKED ISSUE #263: linux_distribution() return on CentOS differs from the README
## From the README

```python
>>> import distro
>>> distro.linux_distribution(full_distribution_name=False)
('centos', '7.1.1503', 'Core')
```

## Actual behavior

```python
>>> import distro
>>> distro.linux_distribution(full_distribution_name=False)
('centos', '7', 'Core')
```

The same goes for CentOS 8:

```python
>>> import distro
>>> distro.linux_distribution(full_distribution_name=False)
('centos', '8', 'Core')
```

## Possible solutions

1. Use `best=True` when getting the version in `linux_distribution()`
2. Add a `best_version` argument to `linux_distribution()` to allow this to be passed through.

I do think that adding a `best_version` argument would be a good idea. However, I'm not sure defaulting to `best=True` helps, since it adds the maintenance releases to Ubuntu LTS releases.

So, perhaps this current behavior is best, since it makes the behavior of `linux_distribution` more congruent with other Linux distros. If that is indeed the position of the project maintainers, and no changes need to be made to the code, I can open a PR to update the README.

terminalmage:
This passes them through as `pretty` and `best` when calling version().

Resolves #263

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,16 @@ $ distro -j
 
 $ python
 >>> import distro
+>>> distro.linux_distribution()
+('CentOS Linux', '7', 'Core')
 >>> distro.linux_distribution(full_distribution_name=False)
-('centos', '7.1.1503', 'Core')
+('centos', '7', 'Core')
+>>> distro.linux_distribution(pretty_version=True)
+('CentOS Linux', '7 (Core)', 'Core')
+>>> distro.linux_distribution(best_version=True)
+('CentOS Linux', '7.7.1908', 'Core')
+>>> distro.linux_distribution(pretty_version=True, best_version=True)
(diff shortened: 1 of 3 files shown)


nir0s:
This is way overdue, but also interesting. I mean.. it's extremely opinionated (regardless of the fact that it's ultra-breaking).

Thanks @terminalmage, but can you clarify why this warrants the breakage?

terminalmage:
Did you read the linked issue?

nir0s:
I have indeed, but I now see that I missed the last part.

I'll comment in the issue.

PR #268, opened 2020-11-11
Ignore file 'iredmail-release'

iredmail:
Ignore file `/etc/iredmail-release` while detecting target OS distribution, it may cause Ansible incorrectly identify target OS distro or release version.

File /etc/iredmail-release is generated by iRedMail - a popular open source mail server solution: https://www.iredmail.org/

- See also: https://github.com/ansible/ansible/pull/72538


sethmlarson (approved):
LGTM, thank you!

PR #269, opened 2021-01-03
Add mypy type checking

jdufresne:
Type checking helps build confidence that an internally consistent API
is used correctly.

Use type comments rather than annotations to retain compatibility with
Python 2.7.

The functools.cached_property is now used on Python 3.8+ as this helps
mypy understand the type of the cached property.

Rationale
---------

The pip project vendors a copy of distro.py:
https://github.com/pypa/pip/blob/20.3.3/src/pip/_vendor/distro.py

pip also uses mypy. Right now, pip must workaround or ignore type errors
that occur from calls to distro.py. By including types directly in the
library, it will help distro, pip, and any other library users running
mypy.

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -37,6 +37,43 @@ import logging
 import argparse
 import subprocess
 
+# Use `if False` to avoid an ImportError on Python 2. After dropping Python 2
+# support, can use typing.TYPE_CHECKING instead. See:
+# https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING
+if False:  # pragma: nocover
+    from typing import (
+        Any,
+        Callable,
+        Dict,
+        Iterable,
+        Optional,
+        Sequence,
+        TextIO,
+        Tuple,
+        Type,
+        TypedDict,
+        Union,
+    )
+
+    VersionDict = TypedDict(
+        'VersionDict',
+        {
+            'major': str,
+            'minor': str,
+            'build_number': str,
+        },
+    )
+    InfoDict = TypedDict(
+        'InfoDict',
+        {
+            'id': str,
+            'version': str,
+            'version_parts': VersionDict,
+            'like': str,
+            'codename': str,
+        },
+    )
+
 
 _UNIXCONFDIR = os.environ.get('UNIXCONFDIR', '/etc')
 _OS_RELEASE_BASENAME = 'os-release'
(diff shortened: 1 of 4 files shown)


distro.py
```diff
@@ -37,6 +37,40 @@
 import argparse
 import subprocess
 
+if False:
```
funkyfuture, line 40: what's the rationale for this condition? is that to work with Python 2, where no `typing.TYPE_CHECKING` is available? if so, a short comment would help in the long term for other readers.

amending `# pragma: nocover` to this line should calm down that automated commenter here.

funkyfuture (commented):
i find these changes very helpful and well done.

distro.py
jdufresne: Thanks for the review @funkyfuture. I have applied your suggestions to the latest revision.

distro.py
```diff
@@ -942,6 +1029,7 @@ def _os_release_info(self):
 
     @staticmethod
     def _parse_os_release_content(lines):
+        # type: (TextIO) -> Dict[str, str]
```
sethmlarson, line 1032: Instead of `TextIO` should this be `Iterable[str]`?

sethmlarson (changes requested):
The change looks good, just need to hook it up to GHA (and one more comment too)

.travis.yml
```diff
@@ -15,6 +15,10 @@ matrix:
       env: TOXENV=py3flake8
```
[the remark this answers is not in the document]
jdufresne: Done in latest :slightly_smiling_face:

distro.py
jdufresne: No, this should remain `TextIO`. This is passed an open file handler. 

The `lines` argument is passed to `shlex` which you can see also requires `TextIO`:

https://github.com/python/typeshed/blob/96e0660fba564d574d9bc69b6b151400d00c7f0a/stdlib/shlex.pyi#L31

Changing this to `Iterable[str]` results in the following failure:

```
distro.py:1047: error: Argument 1 to "shlex" has incompatible type "Iterable[str]"; expected "Union[str, TextIO, None]"
```

sethmlarson: Sounds good, thanks for the explanation!

sethmlarson (approved):
LGTM, thank you!

PR #271, opened 2021-02-13
Use binary file for devnull

methane:
This is not a bug. But binary file is more efficient because TextIOWrapper is not used.


hartwork (commented):
Seems to work. I have tried:
```python
import os
import subprocess

with open(os.devnull, 'wb') as devnull:
    argv = ['sh', '-c', 'echo stdout; echo stderr >&2']
    subprocess.check_output(argv, stderr=devnull)
```

methane:
We can use `stderr=subprocess.DEVNULL` isntead of opening it manually if we can drop Python 2 and 3.2 support.

jdufresne:
> We can use `stderr=subprocess.DEVNULL` isntead of opening it manually if we can drop Python 2 and 3.2 support.

I opened #281 to remove EOL Pythons, allowing for using `DEVULL`

jdufresne (dismissed):

nir0s (approved):

PR #275, opened 2021-06-30
Add approvable workflow to deploy to PyPI

sethmlarson:
Since we're in an organization now, instead of having any individual able to deploy to PyPI (security risk!) I propose we create an approvable workflow using GitHub Actions where (after a tag is created) two maintainers, one initiating and one approving can build and deploy the tag to PyPI.

This requires the following:
- Setting up the Workflow/Environment in GitHub, I'm not sure if it's automatically done as I've not created one before, only used them.
- @nir0s create an API token on PyPI with the `distro` scope and attach it to the `Deploy` environment.
- Setting the branch protection policy to allow writes and tags from at least one developer in the organization.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: "Deploy to PyPI"
+    runs-on: ubuntu-latest
+    environment: Deploy
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ github.event.inputs.tag }}
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
HorlogeSkynet, line 20: Picky hat on : I'd rather not fix Python version here. Always relying on stable should be enough for this need.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,27 @@
...
+        description: "Git tag to deploy to PyPI"
+        required: true
+
+jobs:
+  deploy:
+    name: "Deploy to PyPI"
+    runs-on: ubuntu-latest
+    environment: Deploy
+    steps:
+      - uses: actions/checkout@v1
```
HorlogeSkynet, line 15: Picky hat still on : `actions/checkout@v2` is available.

.github/workflows/deploy.yml
sethmlarson: TIL `3.x` is supported, I've got a couple places I can update now.

.github/workflows/deploy.yml
sethmlarson: Sure

funkyfuture (approved):
lgtm. you could also bump to a dev version that can then be used to test the workflow.

HorlogeSkynet (approved):

nir0s (approved):

nir0s:
@sethmlarson, the API key is in place.

sethmlarson:
Awesome! We can test this with the next release.

PR #277, opened 2021-07-01
Switch CI to GitHub actions

jdufresne:
Since June 15th, 2021, the building on travis-ci.org has ceased.

GitHub Actions also supports Windows, so move away from AppVeyor as
well.


README.md
```diff
@@ -1,8 +1,7 @@
 Distro - an OS platform information API
 =======================================
 
-[![Build Status](https://travis-ci.org/nir0s/distro.svg?branch=master)](https://travis-ci.org/nir0s/distro)
-[![Build status](https://ci.appveyor.com/api/projects/status/e812qjk1gf0f74r5/branch/master?svg=true)](https://ci.appveyor.com/project/nir0s/distro/branch/master)
+[![Build Status](https://github.com/jdufresne/distro/workflows/Build/badge.svg)](https://github.com/jdufresne/distro/actions/workflows/build.yaml)
```
sethmlarson, line 4: We should probably use `github.com/python-distro/distro`?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,93 @@
+name: Build
```
sethmlarson, line 1: Since this is CI and not a "build" per-say, should we use `ci.yml` and `name: CI`?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install tox
+        run: python -m pip install tox
+
+      - name: Run
```
sethmlarson, line 31: `Run tox`?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,93 @@
...
```
sethmlarson, line 67: same as above

sethmlarson (changes requested):
This looks great, will require @nir0s to disable Travis and change the branch protection. Bunch of comments for you though:

README.md
jdufresne: Thanks for catching this. I had intended to go back and fix after testing. Fixed now in latest.

.github/workflows/build.yaml
jdufresne: Works for me. Updated in latest.

.github/workflows/build.yaml
jdufresne: :+1:

jdufresne:
Thanks for the review @sethmlarson! All feedback applied in the latest revision.

sethmlarson (approved):
LGTM

nir0s (approved):

PR #278, opened 2021-07-01
Add CODEOWNERS

nir0s:
Could benefit us by focusing PR notifications on maintainers

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @python-distro/maintainers


nir0s:
Thoughts?

sethmlarson:
I'm eh on a global policy for pings, I'm already following the repository after all :) Usually I use `CODEOWNERS` to protect/solicit reviews from a select group of individuals rather than "everyone".

nir0s:
Yeah, though that's what usually frightens me. Not everyone who joins an OS org is a "maintainer" per-se. Then you get reviews where you don't want them.

nir0s:
I guess I'm going to close this until we feel the pain :)

PR #279, opened 2021-07-01
Add stale.yml config

nir0s:
I've started using staling configs across many projects. I think it keeps things clean. It may be a tad controversial in such open projects, so I'd like to get opinions here.


sethmlarson (approved):
I've never used a stale service before but it seems useful, especially with the pretty lax config here. Seems good to me :)

nir0s:
Right, so at Strigo, we have 90d for issues and 7d for PRs. It works wonderfully. Here, it won't work 😊

PR #280, opened 2021-07-01
Replace flake8 with pylint

nir0s:
* Add `.pylintrc`
* Add pylint config to CI workflow and tox.
* Fix linting issues. For more info, see the commit message for https://github.com/python-distro/distro/commit/85c52b85c7ca8ec02770d04c90eb9cd4527b410a
* Remove py27 linting. It's EOL. Not sure if that makes sense right now, given that we test on 27. However, I'm not sure we should lint for it. Thoughts?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,8 +12,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - python: 2.7
-            toxenv: pylint27
           - python: 3.8
             toxenv: pylint38
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,9 +13,9 @@ jobs:
       matrix:
         include:
           - python: 2.7
-            toxenv: flake8
+            toxenv: pylint27
           - python: 3.8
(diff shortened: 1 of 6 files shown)


jdufresne:
> Remove py27 linting. It's EOL. Not sure if that makes sense right now, given that we test on 27. However, I'm not sure we should lint for it. Thoughts?

I think we should remove EOL Pythons. I've done so in PR #281.

sethmlarson:
Isn't flake8 more complete than pylint? I've not seen projects moving from flake8 -> pylint, usually the other way.

jdufresne:
FWIW, most projects I have worked on choose flake8 over pylint. So it is my personal choice as well. (But only that, personal)

hartwork:
@nir0s what is the motivation behind choosing pylint over flake8?

nir0s:
Yeah, I was making assumptions here. I personally moved away from flake8 in different projects because I felt like it has better reporting.

However, I'm not strongly opinionated about this.

nir0s:
I'm gonna close it. It's not worth a discussion AFAIC.

PR #281, opened 2021-07-01
Drop support for EOL Pythons

jdufresne:
Python 2.7 and 3.4, and 3.5 are end of life. They are no longer
receiving bug fixes, including for security issues. For additional
details on support Python versions, see:

Supported: https://devguide.python.org/#status-of-python-branches
EOL: https://devguide.python.org/devcycle/#end-of-life-branches

Removing support for EOL Pythons will reduce testing and maintenance
resources while allowing the library to move towards a modern Python 3
style.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,8 +21,6 @@ classifiers =
     Operating System :: POSIX :: BSD :: OpenBSD
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.4
-    Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
@@ -32,6 +30,7 @@ classifiers =
 
 [options]
 py_modules = distro
+python_requires = >=3.6
 
 [options.entry_points]
 console_scripts =
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,9 +19,8 @@ classifiers =
     Operating System :: POSIX :: BSD :: FreeBSD
     Operating System :: POSIX :: BSD :: NetBSD
     Operating System :: POSIX :: BSD :: OpenBSD
-    Programming Language :: Python :: 2
-    Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.4
     Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
@@ -37,9 +36,6 @@ py_modules = distro
 [options.entry_points]
 console_scripts =
     distro = distro:main
-
-[bdist_wheel]
-universal = 1
 
 [flake8]
 max-line-length = 88
--- a/tests/test_distro.py
+++ b/tests/test_distro.py
@@ -1,4 +1,3 @@
-# coding=utf-8
 # Copyright 2015,2016 Nir Cohen
 #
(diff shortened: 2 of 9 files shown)


nir0s:
Well, I think that it would be nice to support py27 in a branch like we do with 26, and then drop support in master.

We can have maintainers for these branches. WDYT?

docs/conf.py
```diff
@@ -336,7 +334,7 @@ def parse_version():
 # to. Example RST source for 'py' prefix:
 #     :py:func:`platform.dist`
 intersphinx_mapping = {
-  'py': ('https://docs.python.org/3.5', None)
+  'py': ('https://docs.python.org/3', None)
```
sethmlarson, line 337: This change would be useful outside of this PR.

jdufresne:
> Well, I think that it would be nice to support py27 in a branch like we do with 26, and then drop support in master.
> 
> We can have maintainers for these branches. WDYT?

Works for me. Could someone with permission create a new branch?

docs/conf.py
jdufresne: :+1: Done in #282

sethmlarson:
If we're going the way of removing Python 2 support are we planning a major release to signal this? Then the branch name could be `1.x`?

jdufresne:
> If we're going the way of removing Python 2 support are we planning a major release to signal this? Then the branch name could be `1.x`?

This all makes sense to me and is a good idea.

sethmlarson:
@jdufresne This PR might be hanging out for a while, perhaps we should open an issue so we can discuss what needs to happen outside of code changes to safely drop Python 2?

jdufresne:
> @jdufresne This PR might be hanging out for a while, perhaps we should open an issue so we can discuss what needs to happen outside of code changes to safely drop Python 2? After we get all that out of the way we can then merge this, but it might get out of date in the meantime.

Good idea. I opened issue #291. Please feel free to add to the checklist if you see other necessary steps.

tests/test_distro.py
```diff
@@ -741,7 +736,7 @@ def _test_outcome(
             DISTROS_DIR,
             distro_name + version,
             "etc",
-            "{0}-{1}".format(release_file_id, release_file_suffix),
+            "{}-{}".format(release_file_id, release_file_suffix),
```
HorlogeSkynet, line 739: What about an F-string here too @jdufresne ?

```suggestion
            f"{release_file_id}-{release_file_suffix}",
```

jdufresne: Thanks for the suggestion. I've converted this to an f-string in the latest revision. I also converted a few other discovered by the tool [flynt](https://github.com/ikamensh/flynt).

setup.cfg
```diff
@@ -19,9 +19,8 @@ classifiers =
     Operating System :: POSIX :: BSD :: FreeBSD
     Operating System :: POSIX :: BSD :: NetBSD
     Operating System :: POSIX :: BSD :: OpenBSD
-    Programming Language :: Python :: 2
-    Programming Language :: Python :: 2.7
     Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.4
```
sethmlarson, line 24: Need to remove the Python :: 3.4 and 3.5 as well.

sethmlarson (approved):
Apart from my one comment this change looks good to me :) Thank you!

setup.cfg
jdufresne: Good catch! Fixed in latest revision, commit "Fixup setup.cfg".

sethmlarson (approved):
LGTM, thank you for this awesome PR :muscle:

PR #284, opened 2021-07-01
Add pre-commit config

nir0s:
Would be nice to save the trouble of having to run pylint on every push.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,22 @@
+repos:
+- repo: local
+  hooks:
+  - id: pylint
+    name: pylint
+    # Don't wanna deal with tests right now.
+    entry: pylint --ignore-patterns=test_.*?py
+    args: ["--score=n"]
+    language: python
+    types: [python]
+  - id: black
+    name: black
+    entry: black --target-version py27 --check --diff .
+    language: python
+    require_serial: true
+    types: [python]
+  - id: mypy
+    name: mypy
+    entry: mypy --strict distro.py
+    language: python
+    require_serial: true
+    types: [python]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,22 +1,20 @@
 repos:
-- repo: local
-  hooks:
-  - id: pylint
-    name: pylint
-    # Don't wanna deal with tests right now.
-    entry: pylint --ignore-patterns=test_.*?py
-    args: ["--score=n"]
-    language: python
-    types: [python]
-  - id: black
-    name: black
-    entry: black --target-version py27 --check --diff .
-    language: python
-    require_serial: true
-    types: [python]
-  - id: mypy
-    name: mypy
-    entry: mypy --strict distro.py
-    language: python
-    require_serial: true
-    types: [python]
+  - repo: https://github.com/psf/black
+    rev: 21.6b0
+    hooks:
+      - id: black
+        args: ["--target-version", "py27"]
+  - repo: https://github.com/PyCQA/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+  - repo: https://github.com/pycqa/isort
+    rev: 5.9.1
+    hooks:
+      - id: isort
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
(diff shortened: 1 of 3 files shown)


jdufresne:
pre-commit is nice. This could include all the lint tools.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+    entry: pylint --ignore-patterns=test_.*?py
+    args: ["--score=n"]
+    language: python
+    types: [python]
+  - id: black
+    name: black
+    entry: black --target-version py27 --check --diff .
+    language: python
+    require_serial: true
+    types: [python]
```
jdufresne, line 16: Can simplify

```suggestion
  - repo: https://github.com/psf/black
    rev: 21.6b0
    hooks:
      - id: black
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+repos:
+- repo: local
+  hooks:
+  - id: pylint
+    name: pylint
+    # Don't wanna deal with tests right now.
+    entry: pylint --ignore-patterns=test_.*?py
+    args: ["--score=n"]
+    language: python
+    types: [python]
```
jdufresne, line 10:
```suggestion
  - repo: https://github.com/PyCQA/flake8
    rev: 3.9.2
    hooks:
      - id: flake8
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,22 @@
+repos:
+- repo: local
```
jdufresne, line 2: We can leverage the pre-commit configs of other repos to avoid duplicating their configuration.

jdufresne (commented):
This could add isort too

```
  - repo: https://github.com/pycqa/isort
    rev: 5.9.1
    hooks:
      - id: isort
```

.pre-commit-config.yaml
nir0s: My experience with not using local is that it always screws things up somehow. However, I haven't tried this in a while. I suggest we merge and then try out using remote configs.

WDYT?

jdufresne: > My experience with not using local is that it always screws things up somehow.

I haven't bumped into that, do you have more details or an example?

I think we should go with the more conventional documented approach -- which is using the library provided configs -- unless we can demonstrate an issue with them.

jdufresne:
Here is an example that uses remote repos. This works for me locally. I'd be interested if you bump into the problem you've seen in the past.

This also makes `tox.ini` go through the pre-commit to run lint so all environments are consistent.

```diff
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
new file mode 100644
index 0000000..ef9f075
--- /dev/null
+++ b/.pre-commit-config.yaml
@@ -0,0 +1,20 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 21.6b0
+    hooks:
+      - id: black
+        args: ["--target-version", "py27"]
+  - repo: https://github.com/PyCQA/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+  - repo: https://github.com/pycqa/isort
+    rev: 5.9.1
+    hooks:
+      - id: isort
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.910
+    hooks:
+      - id: mypy
+        args: ["--strict"]
+        files: ^distro\.py$
diff --git a/tox.ini b/tox.ini
index 4ca8df5..e207cb5 100644
--- a/tox.ini
+++ b/tox.ini
@@ -32,13 +32,6 @@ basepython = {env:PYTHON:}\python.exe
 passenv= ProgramFiles LOGNAME USER LNAME USERNAME HOME USERPROFILE
 
 [testenv:lint]
-deps =
-    black
-    flake8
-    isort
-    mypy
-commands =
-    black --target-version py27 --check --diff .
-    flake8
-    isort --check --diff .
-    mypy --strict distro.py
+deps = pre-commit
+commands = pre-commit run --all-files --show-diff-on-failure
+skip_install = true
```

.pre-commit-config.yaml
nir0s: nm. Updated as per your suggestion.

jdufresne (approved):

PR #290, opened 2021-07-01
Introduce Black for automatic code formatting

jdufresne:
As you may know, Black is an opinionated code formatter. By using it,
all contributions will use a single consistent code formatting style.
Let the tools and CI do the mundane formatting work so contributions and
reviewers don't have to. Using an automatic code formatting can also
reduce review friction as there are fewer personal choices to discuss.

Black is frequently used by many projects in the Python community. So
using it will also make the distro code look more like other Python
code.

For additional details on Black, see:
https://github.com/psf/black


hartwork (commented):
Just my two cents: I personally consider the output of black often so ugly, that I went for [yapf](https://github.com/google/yapf) instead in other projects, to get the same benefits while keeping code well readable by humans, not just machines.

Here's an example of black versus yapf on the same code base, where we went with yapf rather than black:
- black — https://github.com/git-big-picture/git-big-picture/pull/56 
- yapf — https://github.com/git-big-picture/git-big-picture/pull/66

sethmlarson (approved):
I'm in favor of this change. I only found one instance in two places where readability was degraded, otherwise readability was kept the same or improved.

jdufresne:
Thanks for the feedback @hartwork 

At one time I too considered Black output sub-optimal, but I have come to realize that my initial reaction was really just about seeing something different, but not necessarily ugly. Since then, I have used Black in multiple projects. It is now very familiar to me and so I have come to find its output predictable and pleasant. Perhaps if you gave it time, you might feel the same way, but for all I know, you already gave it a fair shake.

Even if we still find some parts of it sub-optimal, I think having a single choice and a single tool is ultimately the better choice.

distro.py
```diff
@@ -84,9 +79,7 @@
 #:   with blanks translated to underscores.
 #:
 #: * Value: Normalized value.
-NORMALIZED_OS_ID = {
-    'ol': 'oracle',  # Oracle Linux
-}
+NORMALIZED_OS_ID = {"ol": "oracle"}  # Oracle Linux
```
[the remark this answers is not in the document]
jdufresne: This can be fixed with "magic trailing commas" I will do that.

jdufresne:
@sethmlarson Those identified cases have been restored in the latest revision.

sethmlarson:
The magic trailing comments worked, this is now 100% an improvement to the codebase IMO. I'm going to hold off on merging until we can have consensus.

Adding to @jdufresne, Black has come a long way since it's first introduction and now has an additional mission of "reduce lines modified per diff" that it fulfills very well compared to other style tool's mission of "make the code look consistent" and for this reason I favor it over other tools.

hartwork:
@sethmlarson if everyone else wants Black, please go with Black. I understand the mission of Black but "consistently ugly" is not a gain in my book, because I value readability over consistency. I understand that it's just my opinion.

nir0s:
I'm with @hartwork. I'd go with yapf.

sethmlarson:
Sounds like yapf is the choice, any automatic formatter is better than none.

jdufresne:
I see 2 for black and 2 for yapf, so no clear winner (unless @sethmlarson, you're changing your preference).

We all value readability.

I looked through this diff and I have a hard time understanding what in particular some find "ugly". If you do find Black "ugly", perhaps you could point to a specific line and explain your reasons. I just don't see it. If there is a line to point to, lets make sure there is something substantial and not just a minor preference.

sethmlarson:
@jdufresne I gave favor to @nir0s as author for their choice.

Definitely agree that it'd be good to point out where in this diff exactly isn't working for everyone, I similarly didn't find any changes made by Black to be questionable / "ugly".

nir0s:
Honestly, I think that the best thing to do is to simply use whichever
formatter, and then change it if it bothers anyone, instead of
philosophizing over readability, as it's always an Android vs. iPhone -
like discussion.

I'd merge this and.. let's see how it goes.

hartwork:
> If there is a line to point to, lets make sure there is something substantial and not just a minor preference.

I'm not sure if it's a good idea to go down that rabbit hole: We'll only find that readability is not 100% objective.

To not evade your question though, take `LinuxDistribution.__init__` to demonstrate two key differences:

```python
    # yapf
    def __init__(self,
                 include_lsb=True,
                 os_release_file='',
                 distro_release_file='',
                 include_uname=True,
                 root_dir=None):
        pass

    # black
    def __init__(
        self,
        include_lsb=True,
        os_release_file="",
        distro_release_file="",
        include_uname=True,
        root_dir=None,
    ):
        pass
```

nir0s:
@hartwork, I'm merging this so that we have something running. We can then open an issue and discuss, or PR something that replaces with yapf and debate.

tox.ini
```diff
@@ -33,10 +33,12 @@ passenv= ProgramFiles LOGNAME USER LNAME USERNAME HOME USERPROFILE
 
 [testenv:lint]
 deps =
+    black
     flake8
     isort
     mypy
 commands =
+    black --target-version py27 --check --diff .
```
nir0s, line 41: Oh, I actually missed this. Why `py27`?

nir0s:
btw, https://github.com/python-distro/distro/pull/284 adds pre-commit config, to which I'd also add blackening.

tox.ini
hartwork: @nir0s I think because Python 2.7 is still supported and hence no language features that need more recent versions of Python can be used. @jdufresne ?

jdufresne: Yup. Once EOL Pythons are dropped, this can be bumped to `py36`.

PR #292, opened 2021-07-01
Replace setup.py with declarative setup.cfg

jdufresne:
Avoid mixing code with configuration.

Running setuptools no longer runs arbitrary custom code, it is all
driven through a key/value configuration file, setup.cfg.

This also simplifies other tools interaction. For example, the Sphinx
config no longer needs to parse source code (and hope the format hasn't
changed). Now, it can use the configparser library to pull values from
setup.cfg.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,13 +13,14 @@
 # serve to show the default.
 
 import os
-import re
 import sys
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath(".."))
+
+import distro  # noqa: E402
 
 # -- General configuration ------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -12,8 +12,8 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import configparser
```
hartwork, line 15: I think this will need (a) dropping Python 2.7 or (b) use of six or (c) two different imports for Python 2.7 and 3.x to work, see the note on top of https://docs.python.org/2.7/library/configparser.html .

jdufresne: This is for building docs, not running the library. We can make the docs depend on Python 3. There isn't really a need to continue building docs on Python 2.

hartwork: Interesting point. I'll trust your evaluation.

nir0s (dismissed):

setup.cfg
```diff
@@ -1,3 +1,43 @@
+[metadata]
+name = distro
+version = 1.5.0
```
sethmlarson, line 3: Thoughts on closing https://github.com/python-distro/distro/issues/265 by using `version = attr: distro.__version__`?

jdufresne: I addressed the issue in PR #294.

We can't use `version = attr: distro.__version__` as that would break the Sphinx configuration which parses `setup.cfg` to grab the version. I've taken the alternative approach of querying the installed version from `distro.py`. This is safer as other tools can continue to rely on setup.cfg as a source of truth.

jdufresne: @sethmlarson I've applied this suggestion in the latest revision. 

@hartwork This also fixes your suggestion from earlier. The docs can now be built with Python 2 again.

More review feedback welcome.

sethmlarson (approved):
LGTM, thanks for this.

PR #294, opened 2021-07-01
Add distro.__version__attribute

### LINKED ISSUE #265: Module 'distro' has no '__version__' member
As it's case with most Python packages, if you run i.e:

```
print("psutil version:", psutil.__version__)
```

you'll end up getting package version, i.e: `psutil: 5.7.2`

However, if I do the same for distro package, i.e:

```
print("distro:", distro.__version__)
```

I'll immediately get `Module 'distro' has no '__version__' member`

Would it be possible to add this functionality in one of the future versions instead of having to resort to hacks/workarounds in order to print distro package version as part of a Python app.

Thanks!

jdufresne:
Use the version supplied to setuptools as a single source of truth. On
newer Pythons, can use importlib.metadata to acquire the version.

https://docs.python.org/3/library/importlib.metadata.html

Fixes #265

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -36,6 +36,17 @@ import shlex
 import subprocess
 import sys
 
+try:
+    import importlib.metadata
+
+    __version__ = importlib.metadata.version("distro")
+except ImportError:
+    # Python 3.7
+    import pkg_resources  # type: ignore
+
+    __version__ = pkg_resources.get_distribution("distro").version
+
+
 # Use `if False` to avoid an ImportError on Python 2. After dropping Python 2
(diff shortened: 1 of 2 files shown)


HorlogeSkynet:
Maybe `__name__` instead of "distro" redundant litterals ? @jdufresne

distro.py
```diff
@@ -36,6 +36,17 @@
 import subprocess
 import sys
 
+try:
```
sethmlarson, line 39: This won't work if a user copy-pastes the `distro.py` file into their project which I believe is something we support. Maybe this is another reason to have `distro.__version__` be the source of the package version?

PR #296, opened 2021-07-04
Formally deprecate distro.linux_distribution()

jdufresne:
Users should use distro.id(), distro.version(), and distro.name()
instead.

Refs #263

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -35,6 +35,7 @@ import re
 import shlex
 import subprocess
 import sys
+import warnings
 
 __version__ = "1.5.0"
 
@@ -135,6 +136,13 @@ _DISTRO_RELEASE_IGNORE_BASENAMES = (
 def linux_distribution(full_distribution_name=True):
     # type: (bool) -> Tuple[str, str, str]
     """
+    .. deprecated:: 1.6.0
+
+        :func:`distro.linux_distribution()` is deprecated. It should only be
+        used as a compatibility shim with Python's
+        :py:func:`platform.linux_distribution()`. Please use :func:`distro.id`,
+        :func:`distro.version` and :func:`distro.name` instead.
+
     Return information about the current OS distribution as a tuple
     ``(id_name, version, codename)`` with items as follows:
 
@@ -158,6 +166,13 @@ def linux_distribution(full_distribution_name=True):
     method normalizes the distro ID string to a reliable machine-readable value
     for a number of popular OS distributions.
     """
+    warnings.warn(
+        "distro.linux_distribution() is deprecated. It should only be used as a "
+        "compatibility shim with Python's platform.linux_distribution(). Please use "
+        "distro.id(), distro.version() and distro.name() instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
     return _distro.linux_distribution(full_distribution_name)
 
 
--- a/tests/test_distro.py
+++ b/tests/test_distro.py
@@ -2055,9 +2055,11 @@ class TestGlobal:
             assert method_result == function_result
 
         kwargs = {"full_distribution_name": True}
-        _test_consistency("linux_distribution", kwargs)
+        with pytest.deprecated_call():
+            _test_consistency("linux_distribution", kwargs)
         kwargs = {"full_distribution_name": False}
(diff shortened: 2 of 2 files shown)


HorlogeSkynet:
There is a typo : s/compatability/compatibility/.

sethmlarson (commented):
Looks good, one nit:

jdufresne:
Thanks for the reviews. I addressed all feedback in the latest revision.

sethmlarson (approved):
LGTM, pinging @nir0s as this is a big manifesto change. After this should we remove mentions to `distro.linux_distribution()` in the README, repository description, etc

HorlogeSkynet:
Agreeing with @sethmlarson here. From my PoV this project (first) intended to be a drop-in replacement for the `platform` deprecated method. Maybe the official Python documentation should be updated too, as it points to this module for this specific purpose ?

jdufresne:
I'll hold off on merging until @nir0s has a chance to weigh in.

nir0s:
Guys, I appreciate you holding off, but let's not make a habit out of it. If you feel like there's a huge change, and you want me to pitch in, I'll try. However, if you think that it makes sense to do something, and I'm unresponsive, go ahead and do it.

Anyway, to address the comments:

> LGTM, pinging @nir0s as this is a big manifesto change. After this should we remove mentions to distro.linux_distribution() in the README, repository description, etc

Indeed we should.

Agreeing 

> with @sethmlarson here. From my PoV this project (first) intended to be a drop-in replacement for the platform deprecated method. Maybe the official Python documentation should be updated too, as it points to this module for this specific purpose?

It's only my opinion, but I see no reason why there's a need for a single function that returns a tuple with an unclear structure. What's the first part? Is it the full name? The same questions apply to the codename and the version. Given that explicit is better than implicit, I'd rather do something like:

```python
if distro.id() == "..." and distro.name() == "...":
   ...
```

And not

I indeed think that the official Python documentation should be updated.

nir0s:
Guys, I appreciate you holding off, but let's not make a habit out of it. If you feel like there's a huge change, and you want me to pitch in, I'll try. However, if you think that it makes sense to do something, and I'm unresponsive, go ahead and do it.

Anyway, to address the comments:

> LGTM, pinging @nir0s as this is a big manifesto change. After this should we remove mentions to distro.linux_distribution() in the README, repository description, etc

Indeed we should.

> with @sethmlarson here. From my PoV this project (first) intended to be a drop-in replacement for the platform deprecated method. Maybe the official Python documentation should be updated too, as it points to this module for this specific purpose?

It's only my opinion, but I see no reason why there's a need for a single function that returns a tuple with an unclear structure. What's the first part? Is it the full name? The same questions apply to the codename and the version. Given that explicit is better than implicit, I'd rather do something like:

```python
if distro.id() == "..." and distro.name() == "...":
   ...
```

which also allows me to pass fields explicitly to the relevant function (e.g. `best`),

Instead of:

```python
distro_info = distro.linux_distribution()

if distro_info[0] == "you get the point":
  ...
```

I indeed think that the official Python documentation should be updated, and I don't mind deprecating this until distro v2.0 comes out, if only to give people a shitload of time to stop using it.

Honestly, I'm not strongly opinionated about this. `distro` is supposed to provide a reasonable alternative to an annoying problem. If the Python community responds badly to removing `linux_distribution`, AFAIC, we can leave it be. It's not like it takes a lot of effort to maintain.

However, even if we do decide to keep it, we can push to update the official Python docs to encourage people to use `distro.id()`, etc.. instead.

Thoughts?

jdufresne:
Thanks. 

> I indeed think that the official Python documentation should be updated, and I don't mind deprecating this until distro v2.0 comes out, if only to give people a shitload of time to stop using it.

This makes sense. The warning is a good start and there is no rush to remove or anything like that.

HorlogeSkynet:
@python-distro/maintainers 

Dear maintainers, I guess we should update the GitHub repository description now that this deprecation has been released.

Any idea ? What about ?

```diff
- A much more elaborate, renewed alternative implementation for Python's platform.linux_distribution()
+ A much more elaborate replacement for old Python's `platform.linux_distribution()` method
```

Thanks 🙏

PR #297, opened 2021-07-04
Update platform.linux_distribution links to Python 3.7

jdufresne:
As platform.linux_distribution() and platform.dist() were removed from
Python 3, the doc links no longer exist. For these deprecated functions,
use Python 3.7, the last version to include them.


sethmlarson:
Hmmm, imo it makes more sense to point people towards Python 3.7 docs instead of 2?

jdufresne:
That works for me. I've done so in the latest revision.

sethmlarson (approved):
LGTM

PR #298, opened 2021-07-04
Fix several warnings when building the docs

jdufresne:
In index.rst, the word BSD had a single star. Remove it as the word
doesn't need emphasizing. Fixes warning:

    docs/index.rst:23: WARNING: Inline emphasis start-string without end-string.

The section "uname command output" doesn't exist, so remove the link.
Fixes warning:

    docs/index.rst:57: WARNING: Unknown target name: "uname command output".

In commit 7d30cf20c418e7af9aad08e9c6277c9c4a699f2b the name was changed
from "ld" to "distro". One mention in the docs was missed. Fixes the
warning:

    docs/index.rst:84: WARNING: py:func reference target not found: ld.info

The function distro.uname_info is undocumented. Other docs shouldn't
reference it. Fixes warning:

    distro.py:docstring of distro.LinuxDistribution.uname_info:4: WARNING: py:func reference target not found: distro.uname_info

The function uname_release_attr doesn't exist. It was probably intended
to be uname_attr. However, that function is undocumented and so should
also not be documented. Fixes warning:

    distro.py:docstring of distro.LinuxDistribution.uname_info:4: WARNING: py:func reference target not found: distro.uname_info

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -1032,7 +1032,7 @@ class LinuxDistribution(object):
         Return a single named information item from the uname command
         output data source of the OS distribution.
 
-        For details, see :func:`distro.uname_release_attr`.
+        For details, see :func:`distro.uname_attr`.
         """
         return self._uname_info.get(attribute, "")
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,4 +1,3 @@
-
 .. _distro official repo: https://github.com/python-distro/distro
 .. _distro issue tracker: https://github.com/python-distro/distro/issues
 .. _open issues on missing test data: https://github.com/python-distro/distro/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22>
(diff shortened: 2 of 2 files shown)


distro.py
```diff
@@ -991,8 +991,6 @@ def uname_info(self):
         """
         Return a dictionary containing key-value pairs for the information
         items from the uname command data source of the OS distribution.
-
-        For details, see :func:`distro.uname_info`.
```
funkyfuture, line 995: my question is whether the function `uname_info` shall rather appear in the docs.

distro.py
```diff
@@ -1021,8 +1019,6 @@ def distro_release_attr(self, attribute):
         """
         Return a single named information item from the distro release file
         data source of the OS distribution.
-
-        For details, see :func:`distro.distro_release_attr`.
```
funkyfuture, line 1025: this reference works for me with a local build.

jdufresne:
Thanks for the review @funkyfuture I have applied all feedback in the latest revision.

funkyfuture (approved):
thanks for your efforts!

sethmlarson (approved):
LGTM, thank you!

sethmlarson:
@jdufresne I think there's a way we can configure Sphinx to fail on documentation warnings, perhaps this is an avenue we can explore to avoid introducing more in the future as well?

jdufresne:
> @jdufresne I think there's a way we can configure Sphinx to fail on documentation warnings, perhaps this is an avenue we can explore to avoid introducing more in the future as well?

Yup! Exactly where I was going with this. We should add the `-n` and `-W`  options:

```
  -n                nit-picky mode, warn about all missing references
  -W                turn warnings into errors
```

PR #299, opened 2021-07-04
Add py.typed file to package for PEP-561 compliance

jdufresne:
Per PEP-561, packages that include type information that can be consumed
by other libraries should distribute a py.typed file. This tells mypy
and other tools to use type information shipped with the library.

This requires moving distro from a single module file to a package so
that it can ship data files.

For details on PEP-561, see:
https://www.python.org/dev/peps/pep-0561/

For details on mypy using py.typed, see:
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

### DIFF
--- a/distro/__init__.py
+++ b/distro/__init__.py
@@ -1365,7 +1365,3 @@ def main():
         logger.info("Version: %s", distribution_version)
         distribution_codename = dist.codename()
         logger.info("Codename: %s", distribution_codename)
-
-
-if __name__ == "__main__":
-    main()
--- a/distro/__main__.py
+++ b/distro/__main__.py
@@ -1,0 +1,4 @@
+from . import main
+
+if __name__ == "__main__":
+    main()
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,7 +54,7 @@ jobs:
           python-version: ${{ matrix.python }}
 
       - name: Install tox
-        run: python -m pip install tox
+        run: python -m pip install --upgrade tox setuptools
 
       - name: Run tox
         run: tox -e py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,4 +17,4 @@ repos:
     hooks:
       - id: mypy
         args: ["--strict"]
-        files: ^distro\.py$
+        files: ^distro/.*\.py$
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,10 @@ classifiers =
     Topic :: System :: Operating System
 
 [options]
-py_modules = distro
+packages = distro
+
+[options.package_data]
+* = py.typed
 
 [options.entry_points]
 console_scripts =


distro/__main__.py
```diff
@@ -0,0 +1,4 @@
+from . import main
```
funkyfuture, line 1:
```suggestion
from distro import main
```

distro/__init__.py
```diff
@@ -1365,7 +1365,3 @@ def main():
         logger.info("Version: %s", distribution_version)
         distribution_codename = dist.codename()
         logger.info("Codename: %s", distribution_codename)
-
```
funkyfuture, line 1368: why is this moved to a separate module? wouldn't it be better to define a "console script" entrypoint?

funkyfuture (commented):
thank you. it seems to me there's a little more inside that doesn't relate to PEP-561 directly.

distro/__main__.py
jdufresne: IMO the relative import is a better choice. It avoids conflicts with any other "distro" module/package in the Python path. I'd prefer to keep this as-is unless there is a strong reason against it.

distro/__init__.py
jdufresne: There is a console script. That hasn't changed.

It was moved to a module within a package so that additional "data" files can be installed alongside the package. That data file is the PEP-561 py.typed file.

jdufresne: Further, using `__main__.py` allows the user to run `python -m distro` once installed. This was supported with the module version, so I carried over this feature.

nir0s:
Thanks @jdufresne, though this means that `distro.py` will not be distributable in itself any longer. The idea is that you can inject `distro.py` to machines during e.g. provisioning, and simply run `python distro.py -j` to get the info, and maybe use `jq` to parse it.

I'd rather not break that. Thoughts?

jdufresne:
@nir0s I see your point and it is a good one.

It is a bit unfortunate that the module isn't PEP-561 compliant but not such a big deal. AFAICT, it can't be done with a pure module.

funkyfuture:
then what about

1) increasing the overhead to achieve both, alike

```
distro
  |- __init__.py  # imports public api from distro.distro.__all__
  |- __main__.py  # as proposed in this pr
  |- distro.py  # keeps the '__name__ == "__main__"' idiom, needs to define __all__
  |- py.typed  # as proposed in this pr
```

2) documenting all four intended usages in the README

?!

jdufresne:
> then what about
> 
>     1. increasing the overhead to achieve both, alike
> 
> 
> ```
> distro
>   |- __init__.py  # imports public api from distro.distro.__all__
>   |- __main__.py  # as proposed in this pr
>   |- distro.py  # keeps the '__name__ == "__main__"' idiom, needs to define __all__
>   |- py.typed  # as proposed in this pr
> ```
> 
>     1. documenting all four intended usages in the README
> 
> 
> ?!

Thank you for this suggestion and sorry for taking so long to get back to it.

I've implemented this in #315 along with the documentation.

PR #301, opened 2021-07-10
Release v1.6.0

sethmlarson:
Created a changelog with I believe everything that's landed since last release.
[You can take a peek if I missed anything](https://github.com/python-distro/distro/compare/v1.5.0...master). Put the release date on 2021-07-14 to give time to review but if we need to update the date we can on release day.

Once we merge this I'll create a tag and start the deployment which will also require an approval from someone on the team :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,13 +23,6 @@ RELEASE:
 * Moved repository from `nir0s/distro` to
   [`python-distro/distro`](https://github.com/python-distro/distro) on GitHub.
 
-TESTS:
-* Changed CI to use GitHub Actions [[#277](https://github.com/python-distro/distro/pull/277)]
-
-DOCS:
-* Pinned intersphinx links to Python 3.7 [[#297](https://github.com/python-distro/distro/pull/297)]
-
-
 ## 1.5.0 (2020.3.30)
 
 BACKWARD COMPATIBILITY:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,35 @@
+## 1.6.0 (2021.7.14)
+
+BACKWARDS COMPATIBILITY:
+* Deprecated the `distro.linux_distribution()` function. Use
+  `distro.id()`, `distro.version()` and `distro.name()` instead
+  [[#296](https://github.com/python-distro/distro/pull/296)]
+* Deprecated Python 2.7, 3.4 and 3.5 support. Further releases
+  will only support Python 3.6+ 
+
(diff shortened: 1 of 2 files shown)


hartwork (commented):
@sethmlarson that's a great start. I have created a milestone "1.6.0" using your date suggestion and added that milestone to all pull requests and issues that you mention in the change log here. As a result, anyone can now use the listing at https://github.com/python-distro/distro/pulls?q=is%3Apr+is%3Aclosed to see which pull requests were merged after 2020-03 (release 1.5.0) and do not have milestone 1.6.0 added. All those are likely missing from the change log for 1.6.0.

jdufresne (dismissed):
LGTM. Thanks for writing this up!

CHANGELOG.md
```diff
@@ -1,3 +1,35 @@
...
+
+RELEASE:
+* Moved repository from `nir0s/distro` to
+  [`python-distro/distro`](https://github.com/python-distro/distro) on GitHub.
+
+TESTS:
+* Changed CI to use GitHub Actions [[#277](https://github.com/python-distro/distro/pull/277)]
+
+DOCS:
+* Pinned intersphinx links to Python 3.7 [[#297](https://github.com/python-distro/distro/pull/297)]
```
funkyfuture, line 30: i'm not sure whether these are really needed respectively there are other changes like the introduction of black that seem similar to me, and imo aren't really important in this document to 'end users'. what about including [a hyperlink to the filtered commits](https://github.com/python-distro/distro/compare/v1.5.0...master) for this release?

sethmlarson:
Hey folks, just so you know this hasn't left my mind. I'm going on vacation and will be back on Monday, if someone could get the PR into good shape I'm okay with merging it. Thanks and sorry for the wait!

CHANGELOG.md
sethmlarson: I'm not sure how I feel about filtered commits (especially since anyone interested can do this themselves in the GitHub UI) but I do agree that the docs/tests changes aren't relevant to end-users. What do you think?

jdufresne (approved):

CHANGELOG.md
jdufresne: I agree with @sethmlarson . I don't think we need to provide the link. I prefer a curated set of relevant notes, as has been done here.

funkyfuture (approved):

sethmlarson:
I think we're going to move forward with this, can always fix up the changelog post-release as well. Thanks all for the reviews!

PR #302, opened 2021-07-10
Simplify tox and GitHub actions configuration

jdufresne:
Unify the default testenv and pywin. These two environments don't
require any differences. The previously passed environment variables go
unused.

Only install a minimal set of dependencies required to run the tox
environments. This helps them run a tiny bit faster.

Bump minversion to 1.9 due to using the skip_install configuration
option.

Drop the GitHub actions "toxenv" matrix variable. As tox command is
executed with "-e py", tox will use the default Python which is already
controlled by GHA.

Rename GHA lint to Lint for consistent capitalization.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   lint:
-    name: lint
+    name: Lint
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
LGTM, this is much better. Thanks!

Issue #303, opened 2021-07-27
Add guix in the list of distro IDs


akxs14:
Can you please add 'guix' in the list of recognized Distro IDs? It should point to the GNU GUIX System.

sethmlarson:
Part of adding a system is providing all `/etc/*-release` files and the output of `linux_distribution`, could you provide those? Otherwise this will likely not be completable by our team.

brejoc:
@akxs14 On a related note: Maybe you could even open a PR at [higanworks/os_catalog](https://github.com/higanworks/os_catalog)!? But this is of course completely unrelated to getting Guix supported here. Would just be nice to also have it there.

akxs14:
Thank you very much for the quick response @sethmlarson ! os-release is attached below (I added the extension *.txt so it can be attached in the message). 

[os-release.txt](https://github.com/python-distro/distro/files/6896937/os-release.txt)

The output of `linux_distribution` is  `('GNU', '', '')` . Please let me know if you need anything else.

akxs14:
Very happy to do so @brejoc !

akxs14:
Hi again @sethmlarson , would it help if I open a PR with the requested addition?

sethmlarson:
@akxs14 Yes that'd be awesome!

akxs14:
Ping @sethmlarson , @HorlogeSkynet 

https://github.com/python-distro/distro/pull/308

PR #308, opened 2021-08-07
Added GUIX id

akxs14:
Added a reference for GUIX GNU System, no code changes where needed. The output looks like this:

```json
{
    "codename": "",
    "id": "guix",
    "like": "",
    "version": "5.11.15-gnu",
    "version_parts": {
        "build_number": "15",
        "major": "5",
        "minor": "11"
    }
}
```

Let me know if you need anything else.

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -216,6 +216,7 @@ def id() -> str:
     "netbsd"        NetBSD
     "freebsd"       FreeBSD
     "midnightbsd"   MidnightBSD
+    "guix"          GNU GUIX System
     ==============  =========================================
 
     If you have a need to get distros for reliable IDs added into this set,


HorlogeSkynet:
Hey @akxs14, thanks for properly opening this up.

Two things here from my PoV :

* In #303, you stated "The output of linux_distribution is `('GNU', '', '')`". This means that somehow, some information are missing :
    * I _guess_ name does not work because, as your `os-release.txt` shows, `PRETTY-NAME` is different from `PRETTY_NAME` (is it an upstream issue ?).
    * Although, I don't understand yet why second tuple member (version) is not populated.
* IDs stored in the docstring build a "reliable" maintained list of supported distributions. We haven't discussed it, but I guess this means a dedicated test case should be present for your ID in order to prevent any regressing change to goes live.

Waiting for your feedback, bye.

HorlogeSkynet:
Up @akxs14 !

akxs14:
Thanks a lot!

Sent with [ProtonMail](https://protonmail.com/) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

HorlogeSkynet:
@akxs14 could you please read my [previous comment](https://github.com/python-distro/distro/pull/308#issuecomment-894778826) above ?

HorlogeSkynet:
Please re-open when raised concerns are taken into account.

PR #305, opened 2021-07-29
Fix ignored `include_uname` parameter

HorlogeSkynet:
While working on #264 (leading to #304), I noticed that the `include_uname` parameter is completely being ignored by the code base.
This PR addresses this issue and propose a new test to prevent future regression.

Bye 👋

---

**EDIT** : [A quick GitHub search](https://github.com/search?q=include_uname+-filename%3Adistro.py+language%3APython&type=Code) shows that "no one" (at least on this forge) actually uses this flag.

---

**EDIT 2** : I can backport this patch to the `python2.7-support` branch if you're OK with that.

Closes https://github.com/python-distro/distro/issues/309

PR #310, opened 2021-08-14
Fix/use `include_uname` attribute.

Fizzadar:
Closes: https://github.com/python-distro/distro/issues/309


HorlogeSkynet:
Please see #305 🙏

Fizzadar:
@HorlogeSkynet aha! Much better with tests as well, should have checked first :)

PR #311, opened 2021-10-04
Add support for AIX (closes #241)

### LINKED ISSUE #241: Add support for AIX
Add support for AIX. Currently disto returns the following for a AIX 7.2:
```json
{
    "codename": "",
    "id": "aix",
    "like": "",
    "version": "2",
    "version_parts": {
        "build_number": "",
        "major": "2",
        "minor": ""
    }
}
```
Source could be cmd `oslevel`:
```bash
$ oslevel
7.2.0.0
```
`7` --> major
`2` --> minor

HorlogeSkynet:


HorlogeSkynet:
Up @nir0s @jokurz :bow:

HorlogeSkynet:
Up @nir0s @jokurz 🙏

nir0s (dismissed):
LGTM

hartwork (changes requested):
I'd like to raise concerns that AIX is (Unix but) not Linux and that hence putting AIX bits into class `LinuxDistribution` seems like mis-leading modelling. Also I see a mix of host and target system here, e.g. `sys.platform.startswith` is inspected no matter if `.root_dir` is `/`or not, same for the call to `oslevel`. Just my two cents

HorlogeSkynet:
> I'd like to raise concerns that AIX is (Unix but) not Linux and that hence putting AIX bits into class `LinuxDistribution` seems like mis-leading modelling.

You're right, this seems related to #177 raised questions about this module architecture. At the moment, despite renaming `LinuxDistribution` to `UnixDistribution` (in a backward-compatible way until v2.0.0 of course), I don't have any other idea about how we could add AIX support here (`LinuxDistribution()` being a `distro` singleton and so on...).

> Also I see a mix of host and target system here, e.g. host `sys.platform` is inspected no matter if `.root_dir` is `/`or not, same for the call to `oslevel`. Just my two cents

You were absolutely right, please take a look to the updated patch, it aims at improving host and target system information separation (surprisingly it simplified testing and I think this would improve detection rate OOTB).

Bye, thanks for your time Sebastian :bow:

distro.py
```diff
@@ -718,6 +720,16 @@ def __init__(
         self.usr_lib_dir = (
             os.path.join(root_dir, "usr/lib") if root_dir else _UNIXUSRLIBDIR
         )
+        self.path_env = (
+            ":".join(
+                [
+                    os.path.join(root_dir, path.lstrip(os.sep))
+                    for path in _UNIXPATHENV.split(":")
```
hartwork, line 727: `":"` could be `os.pathsep` here with semantic fit, if you like (two times)

hartwork (commented):
@HorlogeSkynet I see some good ideas here, but I'm unsure if this approach will work without use of a `chroot`-like approach, which I don't see here: adjusting `PATH` alone will not keep the commands we are calling from reading files outside/upwards `.root_dir`. Am I missing something?

HorlogeSkynet:
> I see some good ideas here, but I'm unsure if this approach will work without use of a `chroot`-like approach, which I don't see here: adjusting `PATH` alone will not keep the commands we are calling from reading files outside/upwards `.root_dir`. Am I missing something?

No you don't, I actually did think about this before adapting this patch, but then forgot 🤔
I'll replace this mess by a proper `os.chroot` soon. It will also simplify all `os.path.join` operations performed around `root_dir` attribute. Furthermore, replacing `":"` literal  by `os.pathsep` won't be required anymore.
I'll keep you posted, bye 👋

Note : how could we then support `--root-dir` option with `os.chroot` availability limited to UNIX platforms ?

HorlogeSkynet:
So I'm back Sebastian, latest thoughts on this topic :

* `os.chroot` needs at least `CAP_SYS_CHROOT`, and I don't think we can want `distro` to require _any_ privilege in the future
* Actually, even `chroot`-ing wouldn't solve our issue here. The `--root-dir` option cannot be extended to third binaries execution :
    * What if targeted rootfs architecture is different from the host one ?  
    * What if the binary output is based on running kernel information (see `UNAME(2)` for instance, I don't know how `oslevel` works) ?

As AIX detection would be based on information gathered from `uname`, `--root-dir` usage from a Linux host (for instance) targeting an AIX rootfs would be incompatible.

Where do you think we should go from here ?
I see two options from my PoV :

1. Reverting my previous changes and accounting for `--root-dir` limitations, as long as preventing _any_ sub-process execution when this options is used (a.k.a. not trusting `include_uname` nor `include_lsb` constructor values)
2. Opting for a proper `os.chroot` implementation and warning users if they are not on UNIX **or** running with insufficient privileges.

Bye :wave:

HorlogeSkynet:
Dear Sebastian, dear maintainers,

You will find here an updated branch, with 4 (rebased) commits :

1. The first one "naively" brings support for AIX (as originally done here) ;
2. The second one is actually "breaking", as it removes support for binaries execution when `-r/--root-dir` is specified (rationale below) ;
3. The third one updates documentation about `LinuxDistribution` exceptions that can be raised (a #26 omission) ;
4. The last one adds a proper consideration for `root_dir` public attribute.

---

Rationale about `2` : According to the above discussion, concerns have been raised about `-r/--root-dir` argument incompatibility with target rootfs binaries execution.
Bringing support for AIX emphasized this issue, as calling `uname` might return totally wrong information (running kernel would be probed, not the one from the targeted rootfs).
Moreover, about `oslevel` (and maybe in the future **any** interesting binary), we cannot expect it/them from being even executable (architecture, dynamic libraries, specific target system requirement, ...).
The chroot-based approach mentioned above is a dead-end in our case, as, in addition not to fix all of these issues, it requires system privileges that we cannot expect regular users to have/give to.

A new parameter has been introduced (`include_third_bins`) to control execution of each one of these "third-party" binaries that we might happen to call in the future. I'm not sure about its name though, so any comment would be appreciated.

To conclude, [a very quick GitHub search](https://github.com/search?q=LinuxDistribution+root_dir+-filename%3Adistro.py+language%3APython&type=code) shows(?) that actually no one uses this parameter from API (at least on this forge).

Bye, thanks for your time ! :wave:

tests/resources/distros/aix72/bin/uname
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
```
hartwork, line 2: Any reason for this blank (and the one in file `oslevel` above)?

distro.py
```diff
@@ -738,8 +747,9 @@ def __init__(
                 self.os_release_file = usr_lib_os_release_file
 
         self.distro_release_file = distro_release_file or ""  # updated later
-        self.include_lsb = include_lsb
-        self.include_uname = include_uname
+        self.include_lsb = include_lsb if not root_dir else False
+        self.include_uname = include_uname if not root_dir else False
+        self.include_third_bins = include_third_bins if not root_dir else False
```
hartwork, line 752: I see some conflict here. What if we get passed say `include_uname=True, include_third_bins=False`? Maybe raise `ValueError` with helpful text whenever `(include_lsb or include_uname) and not include_third_bins` holds true?

distro.py
```diff
@@ -749,10 +759,13 @@ def __repr__(self) -> str:
             "distro_release_file={self.distro_release_file!r}, "
             "include_lsb={self.include_lsb!r}, "
             "include_uname={self.include_uname!r}, "
+            "include_third_bins={self.include_third_bins!r}, "
```
hartwork, line 762: You mentioned that `include_third_bins` may not be the best name, and I agree, its a bit weird. I know `allow_subprocesses` is longer and sounds like specific to the subprocess module. In general I think finding another name would be nice.

distro.py
hartwork, line 752: One option for unprivileged execution of binaries from a chroot is [proot](https://proot-me.github.io/). It would only work for binaries from a compatible architecture (e.g. x86 on x86_64 may work or not, depending on kernel config) and any data that app gets from the live kernel would be from the host kernel, not the chroot kernel. And it would need to be installed. Running commands from the chroot without any chroot-like mechanism seems troublesome, yes.

hartwork (commented):
Hi @HorlogeSkynet , good summary, sorry for not getting to a review and reply earlier:

tests/resources/distros/aix72/bin/uname
HorlogeSkynet: Just for consistency with most of other testing shell scripts :wink:

distro.py
HorlogeSkynet: Actually this proposal implied that `include_third_bins` does not "target" `lsb_release` and `uname` calls. See next comment for another possible "fix".

distro.py
HorlogeSkynet: What about `include_oslevel` then ?
I tried an _abstract_ approach for any "new" binaries we would add support for in the future, but maybe we want to leave the possibility for developers to choose which programs they want to run or not.
(This would also fix the concern raised in your previous comment)

hartwork: Sounds like a good "hack" to evade multiple problems at once :+1:

tests/resources/distros/aix72/bin/uname
hartwork: Okay :smile: :man_shrugging:

src/distro/distro.py
```diff
@@ -738,8 +749,9 @@ def __init__(
                 self.os_release_file = usr_lib_os_release_file
 
         self.distro_release_file = distro_release_file or ""  # updated later
-        self.include_lsb = include_lsb
-        self.include_uname = include_uname
+        self.include_lsb = include_lsb if not root_dir else False
+        self.include_uname = include_uname if not root_dir else False
+        self.include_oslevel = include_oslevel if not root_dir else False
```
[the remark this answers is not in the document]
HorlogeSkynet: Dear Sebastian, my quick experience with `proot` has not been conclusive so I preferred keeping this final patch "simple" with a quick runtime warning.
If you have the time and the will to include `os.chroot`/`proot` proper usages for this issue, feel free to edit this branch, or maybe in the future in another PR.
--> I'm switching on the other branch !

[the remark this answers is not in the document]
hartwork: @HorlogeSkynet I'm starting to think that executing binaries from the chroot could also be unexpected and then even have security implications. Maybe a future PR (and additional parameter) are indeed a good idea.
I'm also wondering if we have protection against symlinks pointing to absolute paths outside the chroot anywhere. Are we "manually" resolving symlinks already? I think we'd need to :thinking:

src/distro/distro.py
```diff
@@ -738,8 +749,22 @@ def __init__(
...
         self.distro_release_file = distro_release_file or ""  # updated later
-        self.include_lsb = include_lsb
-        self.include_uname = include_uname
+
+        is_root_dir_defined = root_dir is not None
+        if is_root_dir_defined and (include_lsb or include_uname or include_oslevel):
+            logging.getLogger(__name__).warning(
+                "including subprocess data sources from specific root_dir may not work"
+                " as expected."
+            )
```
hartwork, line 758: @HorlogeSkynet is there a chance for the subprocesses to return the right data (other than by coincidence) *ever*? If not, can we turn that into a raise of `ValueError`?

HorlogeSkynet: Good point. In "real" environments (a.k.a not against functional tests resources), I don't think so either.
Any idea to replace the error message ?
Would `Including subprocess data sources from specific root_dir is disallowed to prevent misleading information` do the job ?

hartwork: @HorlogeSkynet I'd replace `misleading` by plain `false` personally but otherwise that seems great, thanks! :+1:

src/distro/distro.py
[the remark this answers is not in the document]
HorlogeSkynet: You are definitely right, and yes, no protection is currently enforced.
I fear `--root_dir` is/will be a PITA to maintain 🙄
Do you think we should add some ` os.path.realpath` calls on concatenated paths _here_ (this patch) ?

hartwork (approved):

src/distro/distro.py
hartwork: @HorlogeSkynet thanks for your work and patience! :+1:

src/distro/distro.py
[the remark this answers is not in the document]
hartwork: @HorlogeSkynet I think it doen't need to be handled in here, no. Seems like orthogonal enough of an issue to be handled separately.

PR #315, opened 2021-10-16
Add py.typed file to package for PEP-561 compliance

jdufresne:
Per PEP-561, packages that include type information that can be consumed
by other libraries should distribute a py.typed file. This tells mypy
and other tools to use type information shipped with the library.

This requires moving distro from a single module file to a package so
that it can ship data files. The original distro.py was not altered.
Only the `__init__.py` and `__main__.py` were added to facilitate being used
as a package. By keeping distro.py unaltered, this allows projects to
continue to vendor the file into their project without as before without
any modifications.

For details on PEP-561, see:
https://www.python.org/dev/peps/pep-0561/

For details on mypy using py.typed, see:
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages


tests/test_distro.py
```diff
@@ -33,11 +33,11 @@
 
 IS_LINUX = sys.platform.startswith("linux")
 if IS_LINUX:
-    import distro
+    import distro.distro
```
nir0s, line 36: How about `from distro import distro`?

nir0s (dismissed):
Overall LGTM.

nir0s:
Please resolve conflicts.

I'm also wondering if there are cases in which people already download directly, which will now break.

Maybe there's a way to redirect?

tests/test_distro.py
jdufresne: That works for me and results in a smaller diff. I made this change in the latest revision.

sethmlarson (approved):
Looks good, one small comment:

sethmlarson:
@jdufresne This PR appears nearly complete, just pinging you in case you still intend to land it.

docs/conf.py
```diff
@@ -25,10 +25,10 @@
 
 # The short X.Y version.
 # Note: We use the full version in both cases.
-version = distro.__version__
+version = distro.distro.__version__
 
 # The full version, including alpha/beta/rc tags
-release = distro.__version__
+release = distro.distro.__version__
```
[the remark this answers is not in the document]
jdufresne: Thanks for catching that. You're right.

I've fixed this in the latest revision.

sethmlarson:
@jdufresne Looks like a recent merge introduced a rebase issue, could you handle that @jdufresne?

jdufresne:
@sethmlarson No problem. Rebased.

sethmlarson (approved):
LGTM, thanks @jdufresne!

PR #324, opened 2022-02-06
Specifies UTF-8 encoding when opening os-release/distro release files

### LINKED ISSUE #243: Test suite assumes UTF-8 encoding while the implementation does not
`distro` implementation simply uses the `open` function without specifying the encoding explicitly. As a result, this causes tests that use unicode-specific samples to fail on systems where default encoding is different:

```
  File "(...)/test_distro.py", line 1653, in test_os_release_attr
    self._test_attr('os_release_info', 'os_release_attr')
  File "(...)/test_distro.py", line 1645, in _test_attr
    info = getattr(_distro, info_method)()
  File "(...)/distro/__init__.py", line 853, in os_release_info
    return self._os_release_info
  File "(...)/distro/__init__.py", line 553, in __get__
    ret = obj.__dict__[self._fname] = self._f(obj)
  File "(...)/distro/__init__.py", line 930, in _os_release_info
  File "(...)/distro/__init__.py", line 961, in _parse_os_release_content
    tokens = list(lexer)
  File "<embedded stdlib>/shlex.py", line 295, in __next__
    token = self.get_token()
  File "<embedded stdlib>/shlex.py", line 105, in get_token
    raw = self.read_token()
  File "<embedded stdlib>/shlex.py", line 136, in read_token
    nextchar = self.instream.read(1)
  File "<embedded stdlib>/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 29: ordinal not in range(128)
```

This issue occurs only in Python 3 (Python 3.6.7 to be precise), in Python 2 this seems not to be a problem.

I am not sure what the best course of action here is. It obviously makes sense to let the operating system decide what encoding is used for storing release information. It is also important to test fringe, unicode input. But it would be also nice to be able to run tests on systems with non-standard encoding.

HorlogeSkynet:
This patch fixes execution when system default locale does not support Unicode.
distro actually didn't honor OS-RELEASE(5), telling us to expect UTF-8 encoded strings : <https://www.freedesktop.org/software/systemd/man/os-release.html#Description>

Against Python 3.10 with `warn_default_encoding` enabled flag, this silences `EncodingWarning`.

Closes #243.

---

EDIT : it looks like Frank does not have write access nor is member of @python-distro/maintainers

### DIFF
--- a/distro.py
+++ b/distro.py
@@ -1025,7 +1025,7 @@ class LinuxDistribution:
             A dictionary containing all information items.
         """
         if os.path.isfile(self.os_release_file):
-            with open(self.os_release_file) as release_file:
+            with open(self.os_release_file, encoding="utf-8") as release_file:
                 return self._parse_os_release_content(release_file)
         return {}
 
@@ -1244,7 +1244,7 @@ class LinuxDistribution:
             A dictionary containing all information items.
         """
         try:
-            with open(filepath) as fp:
+            with open(filepath, encoding="utf-8") as fp:
                 # Only parse the first line. For instance, on SLES there
                 # are multiple lines. We don't want them...
                 return self._parse_distro_release_content(fp.readline())


funkyfuture (approved):
thank you!

hartwork (approved):

PR #325, opened 2022-02-12
Restores compatibility shim with `platform.linux_distribution`

### LINKED ISSUE #238: Distro 1.40: possible regression in distro.codename()
Calls to distro.codename() worked as expected in 1.30, but in 1.40 I'm seeing what amounts to truncated output.

Using Python 3.6.7 on Ubuntu 18.04,

1.30     Bionic Beaver
1.40     bionic

From the documentation, I thought that, on most Linux distros at least, `distro.codename()` should produce results identical to `awk -F '[()]' '/VERSION=/ {print $2}' /etc/os-release`. Is this no longer the case?

HorlogeSkynet:
#230 introduced proper support for os-release `VERSION_CODENAME` attribute but broke compatibility shim with (now defunct) `platform.linux_distribution` as the third tuple member (codename) were not taken from `VERSION` os-release attribute anymore.

Closes #238.

---

> EDIT : Sorry for first submission, wrong revision.

### DIFF
--- a/src/distro/distro.py
+++ b/src/distro/distro.py
@@ -1056,12 +1056,10 @@ class LinuxDistribution:
             # stripped, etc.), so the tokens are now either:
             # * variable assignments: var=value
             # * commands or their arguments (not allowed in os-release)
+            # Ignore any tokens that are not variable assignments
             if "=" in token:
                 k, v = token.split("=", 1)
                 props[k.lower()] = v
-            else:
-                # Ignore any tokens that are not variable assignments
-                pass
 
         if "version" in props:
             # extract release codename (if any) from version attribute


src/distro/distro.py
```diff
@@ -1062,6 +1062,13 @@ def _parse_os_release_content(lines: TextIO) -> Dict[str, str]:
                 # Ignore any tokens that are not variable assignments
                 pass
 
+        if "version" in props:
+            # extract release codename from the the version
+            match = re.search(r"(\(\D+\))|,(\s+)?\D+", props["version"])
+            if match:
+                # release codename appears within parentheses.
+                release_codename = match.group().strip(", ()")
```
hartwork, line 1070: The previous effect of plain `.strip()` seems to have been dropped while moving. Is that intended?

src/distro/distro.py
```diff
@@ -1062,6 +1062,13 @@ def _parse_os_release_content(lines: TextIO) -> Dict[str, str]:
                 # Ignore any tokens that are not variable assignments
                 pass
 
+        if "version" in props:
+            # extract release codename from the the version
+            match = re.search(r"(\(\D+\))|,(\s+)?\D+", props["version"])
```
hartwork, line 1067: Since that code was just moved, maybe it's out of scope here?: The second half of the regex pattern — `|,(\s+)?\D+` — seems weird:
- I cannot find any test cases with pattern like that (was looking with `git grep ', *[0-9]\+'`).
- Any match to that second half will violate the "with parentheses" assumption below.
- `(\s+)?` looks like an overcomplicated version of `(\s*)`.

tests/test_distro.py
```diff
@@ -1905,6 +1908,12 @@ def test_linux_distribution(self) -> None:
         i = _distro.linux_distribution()
         assert i == ("Ubuntu", "14.04", "Trusty Tahr")
 
+        _distro = distro.LinuxDistribution(
+            include_lsb=False, os_release_file=self.fedora30_os_release
+        )
+        i = _distro.linux_distribution()
+        assert i == ("Fedora", "30", "Thirty")
```
hartwork, line 1915: I checked back with a Docker container of Fedora 30 and it seems the copy of `/etc/os_release` is no longer up to date. With the data from the live container, we'd get `Container Image` rather than `Thirty` here. I hope that's okay for a "version"? For proof:

```diff
# diff -u0 ./tests/resources/distros/fedora30/etc/os-release{,-docker}
--- ./tests/resources/distros/fedora30/etc/os-release   2022-02-12 16:38:39.255999496 +0100
+++ ./tests/resources/distros/fedora30/etc/os-release-docker    2022-02-12 16:52:56.252442968 +0100
@@ -2 +2 @@
-VERSION="30 (Thirty)"
+VERSION="30 (Container Image)"
@@ -7 +7 @@
-PRETTY_NAME="Fedora 30 (Thirty)"
+PRETTY_NAME="Fedora 30 (Container Image)"
@@ -12 +12 @@
-DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/30/system-administrators-guide/"
+DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f30/system-administrators-guide/"
@@ -19,0 +20,3 @@
+VARIANT="Container Image"
+VARIANT_ID=container
+
```

Maybe Fedora is in the wrong here and not distro, I'm not sure. Feel free to ignore.

src/distro/distro.py
hartwork: To be more precise: `.strip()` drops whitespace other than spaces. If we only ever have space, we'd be good, but `\s` is more than spaces.

sethmlarson:
Are we able to revert https://github.com/python-distro/distro/pull/230 first to ensure we're restoring compatibility and then evaluate what to do next?

hartwork:
> Are we able to revert #230 first to ensure we're restoring compatibility and then evaluate what to do next?

@sethmlarson it's pretty big and in for 2 years. I'm not sure I'd vote for that path. What's the idea?

src/distro/distro.py
HorlogeSkynet: Yes it was, we are not _supposed_ to encounter whitespaces other than spaces in os-release files. While de-duplicating `strip` calls, I preferred expliciting the list of characters we want to drop purposely. I can revert to the previous behavior if you prefer :+1:

src/distro/distro.py
HorlogeSkynet: > Since that code was just moved, maybe it's out of scope here?

I don't think it is : I only moved it two blocks above to let #230 logic properly overwrite `props["codename"]` as expected.
I didn't want to bring an affectation overhead for this fix, so I've de-duplicated it too.

>The second half of the regex pattern — `|,(\s+)?\D+` — seems weird:
> 
>     * I cannot find any test cases with pattern like that (was looking with `git grep ', *[0-9]\+'`).
> 
>     * Any match to that second half will violate the "within parentheses" assumption below.
> 
>     * `(\s+)?` looks like an overcomplicated version of `(\s*)`.

I've thought too, I am OK to improve it while tweaking those lines :+1:

src/distro/distro.py
hartwork: I'm mostly the messager here. If you want to go with all spaces, then `(\s+)?` probably should become `( *)` (a space and an asterisk).

tests/test_distro.py
HorlogeSkynet: I'm not sure we should base our tests on information shipped in container rootfs, they are known not to represent official images (remember #262 !) :smile:

src/distro/distro.py
hartwork: Oh damn, I was reading `\D` for `\d` :man_facepalming: Sorry! Completely different picture now:

```console
# git --no-pager grep 'VERSION.*, *[^0-9]\+' -- tests/resources/distros/
tests/resources/distros/linuxmint17/etc/os-release:VERSION="14.04.3 LTS, Trusty Tahr"
tests/resources/distros/ubuntu14/etc/os-release:VERSION="14.04.3 LTS, Trusty Tahr"
```

src/distro/distro.py
hartwork: Oh man, I mean as much as much as like non-capturing parentheses, I'm not sure this is the best way forward :joy:

tests/test_distro.py
hartwork: If people live in both worlds, we need to support both. I don't know. I thought I should mention it, maybe it doesn't have to be focus here, your call.

HorlogeSkynet: Yes sir, we do.
From my PoV, we need to support all existing formats, but not their content, as long as they respect the format.
This os-release file you found in Fedora Docker container is indeed different, but does not contain unsupported stuff. So I don't think we should address this here :wink:

src/distro/distro.py
HorlogeSkynet: See updated branch ! I opted for two distinct matching groups, so no extra parsing nor non-capturing groups are required :innocent:

src/distro/distro.py
```diff
@@ -1055,13 +1056,17 @@ def _parse_os_release_content(lines: TextIO) -> Dict[str, str]:
...
-            else:
-                # Ignore any tokens that are not variable assignments
-                pass
 
+        if "version" in props:
+            # extract release codename (if any) from version attribute
+            match = re.search(r"\((\D+)\)|,\s*(\D+)", props["version"])
+            if match:
+                release_codename = match.group(1) or match.group(2)
+                props["codename"] = props["release_codename"] = release_codename
```
hartwork, line 1069: @HorlogeSkynet maybe I missed it but it doesn't seem obvious: Do we even need this new field `release_codename` if…
- `props["codename"]` and therefore `_os_release_info["codename"]` and therfore `.codename()` all have the same right value already and
- it's not used anywhere but in `.linux_distribution()`.

Can it be resolved?

HorlogeSkynet: @hartwork Actually the block you pointed out has been moved "up", but also an `elif` has disappeared : the idea of this patch is to "keep" `release_codename` (in OS-RELEASE(5) terms) available for further usages, even if `codename` is getting overwritten by `VERSION_CODENAME` (or its Ubuntu specific sibling).

I'd have preferred doing this parsing in `linux_distribution` method, but we wouldn't have the "raw" `VERSION` attribute value in such a scope.

**TL; DR** : the block has not _just_ been moved, the `release_codename` field "survives" #230 logic.
We can add an empty new line between `if` blocks if you think it would be appropriate.

HorlogeSkynet: Definitely okay Sebastian ! 🎉

HorlogeSkynet: Done ! I also took the liberty to precise a `linux_distribution` docstring bit, so a final proper review is required 😇

hartwork (approved):

PR #327, opened 2022-02-15
Drastically lowers `LinuxDistribution._distro_release_info` complexity

HorlogeSkynet:


src/distro/distro.py
```diff
@@ -1245,39 +1260,31 @@ def _distro_release_info(self) -> Dict[str, str]:
...
+                # The name is always present if the pattern matches.
+                if "name" not in distro_info:
+                    continue
+                self.distro_release_file = filepath
+                break
+            else:
+                return {}
+
+        if match is not None:
+            distro_info["id"] = match.group(1)
```
hartwork, line 1281: How above resolving this if by moving `distro_info["id"] = match.group(1)` right after/before line 1275 above?

src/distro/distro.py
```diff
@@ -1245,39 +1260,31 @@ def _distro_release_info(self) -> Dict[str, str]:
...
+
+        # manually set id for CloudLinux
+        if "cloudlinux" in distro_info.get("name", "").lower():
+            distro_info["id"] = "cloudlinux"
```
hartwork, line 1285: @HorlogeSkynet let's add a comment here that…
- this is for CloudLinux <7, not CloudLinux in general and that
- we're (not overwriting things here but) adding something missing.

My undererstanding is that `"name"` will be the only field set (to the single free-form line from file `/etc/redhat-release`) for CloudLinux 5 and 6 and this adds an ``"id"` field to be "a little less lacking". My references are:

- https://github.com/python-distro/distro/blob/master/tests/resources/distros/cloudlinux6/etc/redhat-release
- https://github.com/python-distro/distro/blob/master/tests/resources/distros/cloudlinux7/etc/os-release
- https://github.com/python-distro/distro/blob/697b2dbb20ab12cb86becf57f062645de723d5ac/src/distro/distro.py#L1326

What do you think?

src/distro/distro.py
```diff
@@ -1245,39 +1260,31 @@ def _distro_release_info(self) -> Dict[str, str]:
...
+                if not match:
+                    continue
+                filepath = os.path.join(self.etc_dir, basename)
+                distro_info = self._parse_distro_release_file(filepath)
+                # The name is always present if the pattern matches.
+                if "name" not in distro_info:
+                    continue
+                self.distro_release_file = filepath
+                break
+            else:
```
hartwork, line 1277: I know that `else` with loops confuses some people. If you like, maybe add an inline comment like `else:  # i.e. the loop never did a "break"` or something. Just an idea.

src/distro/distro.py
HorlogeSkynet: We could have, but we can't be sure `match` exists when `self.distro_release_file` has been specified (line 1250) !
The affectation being protected here targets both of above conditional branches :wink:

src/distro/distro.py
HorlogeSkynet: You're right, I'll add it !

src/distro/distro.py
hartwork: I see, my bad, you're right! :+1:

HorlogeSkynet:
@hartwork This is the diff I've just amended, to simplify your review :innocent:

More simplifications, and your documentation-related comment.

```diff
diff --git a/src/distro/distro.py b/src/distro/distro.py
index 8209256..4d08067 100755
--- a/src/distro/distro.py
+++ b/src/distro/distro.py
@@ -1250,7 +1250,12 @@ class LinuxDistribution:
             match = _DISTRO_RELEASE_BASENAME_PATTERN.match(basename)
         else:
             try:
-                basenames = os.listdir(self.etc_dir)
+                basenames = [
+                    basename
+                    for basename in os.listdir(self.etc_dir)
+                    if basename not in _DISTRO_RELEASE_IGNORE_BASENAMES
+                    and os.path.isfile(os.path.join(self.etc_dir, basename))
+                ]
                 # We sort for repeatability in cases where there are multiple
                 # distro specific files; e.g. CentOS, Oracle, Enterprise all
                 # containing `redhat-release` on top of their own.
@@ -1262,10 +1267,8 @@ class LinuxDistribution:
                 # error is handled in `_parse_distro_release_file()`.
                 basenames = _DISTRO_RELEASE_BASENAMES
             for basename in basenames:
-                if basename in _DISTRO_RELEASE_IGNORE_BASENAMES:
-                    continue
                 match = _DISTRO_RELEASE_BASENAME_PATTERN.match(basename)
-                if not match:
+                if match is None:
                     continue
                 filepath = os.path.join(self.etc_dir, basename)
                 distro_info = self._parse_distro_release_file(filepath)
@@ -1274,13 +1277,13 @@ class LinuxDistribution:
                     continue
                 self.distro_release_file = filepath
                 break
-            else:
+            else:  # the loop didn't "break": no candidate.
                 return {}
 
         if match is not None:
             distro_info["id"] = match.group(1)
 
-        # manually set id for CloudLinux
+        # CloudLinux < 7: manually enrich info with proper id.
         if "cloudlinux" in distro_info.get("name", "").lower():
             distro_info["id"] = "cloudlinux"
 
```

PR #328, opened 2022-02-18
Manually resolve path relatively to `root_dir` to prevent escape

HorlogeSkynet:
Dear Sebastian, dear maintainers,

This PR drafts a first implementation to address concerns raised in #311 review.

Currently, paths are not manually resolved, and by following symbolic links, this _may_ lead to outside specified `root_dir`.

By using `os.path.realpath` and `os.path.commonprefix`, we can verify that important paths are still prefixed by `root_dir`, once  they have been resolved.

---
It turned out enforcing those checks was not as easy as I thought :
* some paths are provided on purpose as parameters, others automatically derived
* some paths are known during instantiation, others will be "later"
* I'm not sure `PermissionError` actually fits, but I was out of option
* I'm not sure the error message itself is relevant either
* New `__prevent_root_dir_escape` function might require a renaming too

At least, the new functional test actually enlightens the issue.

--> A deep review is required, ideas are welcome and commits on this branch too :upside_down_face:

---

Have a nice WE all :bow:


hartwork (changes requested):
Hi @HorlogeSkynet , maybe I'm misunderstanding something, but right now I only see the part that prevents "unjust" escapes from the chroot (from absolute links or from going above the root) but I don't see any code that would "bend" "legit" absolute symlinks back into the chroot.

Let me demo what I think we'd need in some form:

```python
def resolve_chroot_symlink(link_location, root_dir):
    assert os.path.commonprefix([link_location, root_dir]) == root_dir

    while True:
        try:
            resolved = os.readlink(link_location)
        except FileNotFoundError:  # includes case "not a symlink"
            return link_location

        if resolved.startswith('/'):
            # i.e. absolute path (with regard to the chroot)
            # that we need to move back inside the chroot
            resolved = os.path.join(root_dir, resolved.lstrip('/'))
        else:
            # i.e. relative path that we make absolute
            resolved = os.path.join(os.path.dirname(link_location), resolved)

        if os.path.commonprefix([resolved, root_dir]) != root_dir:
            raise Value('ESCAPE DETECTED')  # TODO

        link_location = resolved
```

What do you think?

HorlogeSkynet:
Dear Sebastian, sorry for the delay.
I got your point and understood what you were trying to achieve : I did miss the part about resolving symbolic links relatively to `root_dir`.

So I tweaked your function a bit in order to :
* ease any future Windows integration task (see #177)
* detect symbolic links loop (as `os.path.realpath` does)
* fix `os.readlink` usage on regular files (it actually raises `OSError: "Invalid argument"`)
* fix `os.path.commonprefix` usage by removing redundant "up-level references"

You will see that I opted for links resolution on the verge of calling `open`. This definitely simplifies the constructor.

I've implemented three short test cases to check our new logic (note : relative symbolic links resolving is already tested by regular rootfs under `tests/resources/distros`).

I also took the liberty to update our `.gitignore` file, as my working directory used not to be clean since #315.

Bye ! Thanks for your time :wave:

.gitignore
```diff
@@ -1,60 +1,120 @@
...
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
```
hartwork, line 22: I would like to vote for not doing unrelated changes to `.gitignore` in here but in a dedicated pull request. Also, the leading slashes should probably be kept since the have meaning and make things go faster (because they apply to fewer places). I should add, that personally I try to keep .gitignore files minimal, i.e. I would not add entries before someone actually hits them. What I see below, seems to fall into that bucket.

src/distro/distro.py
```diff
@@ -1078,6 +1082,47 @@ def uname_attr(self, attribute: str) -> str:
         """
         return self._uname_info.get(attribute, "")
 
+    def __resolve_chroot_symlink(self, link_location: str) -> str:
+        # only resolve symlink if root_dir is set and link is under it
+        if (
+            self.root_dir is None
+            or os.path.commonprefix([self.root_dir, link_location]) != self.root_dir
```
hartwork, line 1089: The second half of this condition feels like we can either drop it or raise an exception if it happens. This silent version seems to either be trouble or not needed.

src/distro/distro.py
```diff
@@ -1078,6 +1082,47 @@ def uname_attr(self, attribute: str) -> str:
...
+            resolved = os.path.normpath(resolved)
+
+            # prevent symlink infinite loop
+            if resolved in seen_paths:
+                raise OSError(errno.ELOOP, os.strerror(errno.ELOOP), resolved)
+
+            if os.path.commonprefix([self.root_dir, resolved]) != self.root_dir:
+                raise RuntimeError(
+                    f"{link_location} resolves outside of {self.root_dir}!"
+                )
```
hartwork, line 1121: I'm not sure `RuntimeError` is the best exception to throw. I think we have another flavor of `OSError` here, maybe even `FileNotFoundError`.
PS: My vote for dropping the exclamation mark. If we add one here, all other exceptions could be argued to need one (but probably don't, either).

src/distro/distro.py
```diff
@@ -1078,6 +1082,47 @@ def uname_attr(self, attribute: str) -> str:
...
+                resolved = os.path.splitdrive(resolved)[1].lstrip(os.sep)
+                if os.altsep is not None:
+                    resolved.lstrip(os.altsep)
+                resolved = os.path.join(self.root_dir, resolved)
+            else:
+                # i.e. relative path that we make absolute
+                resolved = os.path.join(os.path.dirname(link_location), resolved)
+
+            # remove redundant up-level references
+            resolved = os.path.normpath(resolved)
```
hartwork, line 1112: My draft at https://github.com/python-distro/distro/pull/328#pullrequestreview-887955351 was avoiding `normpath` due to this sentence in the API docs:

>  This string manipulation may change the meaning of a path that **contains symbolic links**.

Emphasis mine, from https://docs.python.org/3/library/os.path.html#os.path.normpath .

Maybe this line can just be dropped?

src/distro/distro.py
```diff
@@ -1078,6 +1082,47 @@ def uname_attr(self, attribute: str) -> str:
...
+                resolved = os.readlink(link_location)
+            except OSError:  # includes case "not a symlink"
+                return link_location
+
+            if os.path.isabs(resolved):
+                # i.e. absolute path (regarding to the chroot), that we need to
+                # "move" back inside the chroot
+                resolved = os.path.splitdrive(resolved)[1].lstrip(os.sep)
+                if os.altsep is not None:
+                    resolved.lstrip(os.altsep)
```
hartwork, line 1105: This seems to have Windows in mind but then drops the drive part altogether. Also I don't fully understand the idea behind using `os.altsep`. I think this needs at least another comment, and potentially also a bug fix.

src/distro/distro.py
```diff
@@ -1301,10 +1349,12 @@ def _parse_distro_release_file(self, filepath: str) -> Dict[str, str]:
             A dictionary containing all information items.
         """
         try:
-            with open(filepath, encoding="utf-8") as fp:
+            with open(
+                self.__resolve_chroot_symlink(filepath), encoding="utf-8"
+            ) as release_file:
```
hartwork, line 1354: (I would personally revert `release_file` back to `fp` to reduce "noise" in the diff, but just my 2 cents, please feel free to ignore.)

hartwork (changes requested):
Hi @HorlogeSkynet kudos for the new tests and the loop detection! :+1: I believe a few details need more work, please see my comments below

src/distro/distro.py
HorlogeSkynet: You are right, I rolled-back to `PermissionError` (which seems to fit better), and redacted the `!`.

src/distro/distro.py
HorlogeSkynet: You were right, I ignored the documentation warning and that was a mistake. I added a case to triggering this.

src/distro/distro.py
HorlogeSkynet: Yes, the idea was to "universally" remove absolute "terms" from the path, and I assumed we don't (and can't) know how passed paths are expressed.
On Windows, we need to remove any drive part. Additionally, where `os.altsep` is defined ([as on NT](https://github.com/python/cpython/blob/d4e64cd4b0ea431d4e371f9b0a25f6b75a069dc1/Lib/ntpath.py#L16)) we should strip it as well.

src/distro/distro.py
HorlogeSkynet: I've added this condition as this method might be called with paths not in `root_dir`.
For instance with `LinuxDistribution(root_dir="/mnt", os_release_file="/etc/os-release")`, `os_release_file` is honored (as user-provided), and not derived from `root_dir` (see current version of the constructor).
Feel free to tell me that we should rename this method, or change its prototype.

I've also noticed we are calling `os.path.isfile` (which resolves symbolic links...), so I changed the patch to prevent/workaround this.

HorlogeSkynet:
> Hi @HorlogeSkynet kudos for the new tests and the loop detection! +1 I believe a few details need more work, please see my comments below

:pray:
I've answered/resolved your comments Sebastian.
Although, I'm concerned about the CI failure, as tests pass "on my machine".
I think this is due to an issue related to GitHub Actions and symbolic links resolving to directories (I might be wrong).

Waiting for your inputs, thanks for your time :wave:

src/distro/distro.py
```diff
@@ -1078,6 +1084,53 @@ def uname_attr(self, attribute: str) -> str:
...
+                if (
+                    os.path.commonprefix(
+                        [
+                            os.path.realpath(self.root_dir),
+                            os.path.realpath(link_location),
+                        ]
+                    )
+                    != os.path.realpath(self.root_dir)
+                ):
+                    raise PermissionError(
```
hartwork, line 1109: Maybe not the most important thing in the world, but `FileNotFoundError` seems to be a better match here because:

```console
# cd "$(mktemp -d)"
# ln -s no-such-file broken-symlink
# python -c 'open("broken-symlink")' |& tail -n1
FileNotFoundError: [Errno 2] No such file or directory: 'broken-symlink'
```

From the point of view of inside the chroot it's not a permission error. Are you sure you want to go with `PermissionError`?

src/distro/distro.py
hartwork: `/` on Windows is network shares I guess?

Maybe these low-level bits can be extracted to a new method and then the four lines here could become a high-level one-line `resolved = self.__abs_path_join(self.root_dir, resolved)` or similar.

src/distro/distro.py
hartwork: > I've added this condition as this method might be called with paths not in `root_dir`. For instance with `LinuxDistribution(root_dir="/mnt", os_release_file="/etc/os-release")`, `os_release_file` is honored (as user-provided), and not derived from `root_dir` (see current version of the constructor). Feel free to tell me that we should rename this method, or change its prototype.

If the method can be called like that, then the result could be expected to be resolved while this new code would return plain `link_location` without any resolution. I think I'd vote for only calling `__resolve_chroot_symlink` when we have something from within the chroot. It would help the method do one thing only.


> I've also noticed we are calling `os.path.isfile` (which resolves symbolic links...), so I changed the patch to prevent/workaround this.

That's a great catch, but I'm not sure the fix works as intended. I'd think we needed a version of `os.path.isfile` that uses `__resolve_chroot_symlink` internally.

hartwork (dismissed):

src/distro/distro.py
HorlogeSkynet: > `/` on Windows is network shares I guess?

Not necessarily, on PowerShell for instance, you can use `/` as "regular" path separator :shrug: 

> Maybe these low-level bits can be extracted to a new method and then the four lines here could become a high-level one-line `resolved = self.__abs_path_join(self.root_dir, resolved)` or similar.

Okay :pray:

src/distro/distro.py
HorlogeSkynet: > If the method can be called like that, then the result could be expected to be resolved while this new code would return plain `link_location` without any resolution. I think I'd vote for only calling `__resolve_chroot_symlink` when we have something from within the chroot. It would help the method do one thing only.

I started trying to implement this, it's even messier, as it would add many test conditions here and there... If you have an idea on your side, please feel free to edit this branch :roll_eyes:

> That's a great catch, but I'm not sure the fix works as intended. I'd think we needed a version of `os.path.isfile` that uses `__resolve_chroot_symlink` internally.

I couldn't get your point, sorry Sebastian :disappointed:

HorlogeSkynet:
PR's (finally) ready @hartwork !

src/distro/distro.py
```diff
@@ -1089,6 +1095,59 @@ def uname_attr(self, attribute: str) -> str:
...
+        against ``self.root_dir`` if inside the chroot,
+        else just return the original path.
+        We're doing this check at a central place, to making
+        the calling code more readable and to de-duplicate.
+        """
+        if self.root_dir is None:
+            return link_location
+
+        if os.path.commonprefix([self.root_dir, link_location]) != self.root_dir:
+            raise FileNotFoundError
```
hartwork, line 1118: I believe we need…
- to move this code into the `While True` loop below, and
- to use `os.path.commonpath` (not `commonprefix`) of Python >=3.5 instead OR keep using `os.path.commonprefix` but with a copy of `self.root_dir` ending in exactly one trailing instance of `os.sep`, in order to not run into false positives.

src/distro/distro.py
```diff
@@ -1089,6 +1095,59 @@ def uname_attr(self, attribute: str) -> str:
...
+                if os.path.commonprefix(
+                    [
+                        os.path.realpath(self.root_dir),
+                        os.path.realpath(link_location),
+                    ]
+                ) != os.path.realpath(self.root_dir):
+                    # `link_location` resolves outside of `self.root_dir`.
+                    raise FileNotFoundError from None
+
+                return link_location
```
hartwork, line 1134: I see multiple issues here:
- `os.path.realpath` is following symlinks
- case "not a symlink" is assumed but not verified

How about this substitute for the whole try-except block:
```python
if os.path.commonpath([self.root_dir, link_location]) != self.root_dir:
    raise FileNotFoundError

if not os.path.islink(link_location):
    return link_location
    
resolved = os.readlink(link_location)
```
Whether we'd need to still turn all instances of `OSError` into `FileNotFoundError` am not sure.

What do you think?

src/distro/distro.py
```diff
@@ -1265,7 +1328,6 @@ def _distro_release_info(self) -> Dict[str, str]:
                     basename
                     for basename in os.listdir(self.etc_dir)
```
hartwork, line 1329: I believe this access to `self.etc_dir` needs wrapping with `self.__resolve_chroot_symlink_as_needed`.

tests/test_distro.py
```diff
@@ -734,6 +731,34 @@ def test_empty_release(self) -> None:
...
+        )
+        # loading existing file outside of root_dir has been prevented (empty data)
+        self._test_outcome({})
+
+    def test_rootdirnonescape(self) -> None:
+        self.distro = distro.LinuxDistribution(
+            root_dir=os.path.join(TESTDISTROS, "distro", "rootdirnonescape")
+        )
+        desired_outcome = {"id": "rootdirnonescape"}
+        self._test_outcome(desired_outcome)
```
hartwork, line 753: I understand the idea behind the test itself and consider it a good test, but fail to see `rootdirnonescape` as a good descriptive name. Any other ideas for a name?

hartwork (dismissed):
Hello @HorlogeSkynet, sorry for the big delay. I found some remaining issues. Please see my four comments below for details:

tests/test_distro.py
HorlogeSkynet: Definitely, what about `rootdirfollowsymlink` ? 🙂

hartwork: Hi @HorlogeSkynet, I think I was not paying full attention last time and mis-reading "nonescape" as "none escape" rather than "non-escape", which made significantly less sense to me.  Maybe non-escape is not so bad for a name at all, given that the test exercises a case that looks like an escape at first glance and then turns out to be sane, when having a closer look.  Maybe the best fix is to break with the existing naming scheme "test_notveryreadable" (below) at least for new tests and use e.g. "test_root_dir_non_escape" rather than "test_rootdirnonescape" for a name.  I believe readability would beat consistency here.  What do you think?

HorlogeSkynet: All good for me ! 💯 
Thanks @hartwork 🙏

HorlogeSkynet:
Dear @python-distro/maintainers, Dear @hartwork (if you haven't muted notifications from this repository :upside_down_face:),

I've finally taken the time to rework this feature (fix ?).

What has been done :
* `os.path.commonpath` has been preferred over `commonprefix` (first relative path normalization has been added to prevent standard library exception ;
* 100% (99,99%) code coverage (for newly added code) ;
* `os.path.realpath` have been removed from iteration to prevent OS links resolution, but there is one left for _final_ path to detect some escape cases ;
* Three new tests (two for `os_release_file` as well as another one for above chroot escape case) ;
* `root_dir` is resolved, mainly to deal with `..` (as you correctly pointed out Sebastian) ;
* link check before resolution instead of `try..except` (indeed simpler) ;
* `OSError` has been replaced by `FileNotFoundError` (symlinks loop corresponds to a missing file, at least from the caller PoV) ;
* a "normalization" for relative paths has been added, to strip `../` that would legitimately resolve to chroot in a real-world situation.

This branch as also been rebased, and hence now "ready".

Thanks for your time,
Bye :wave:

hartwork:
> Dear @python-distro/maintainers, Dear @hartwork (if you haven't muted notifications from this repository 🙃),

Hi @HorlogeSkynet, back in 2022-08 — soon two years ago — I implemented a solution on branch [`security/prevent_rootfs_escape_2`](https://github.com/python-distro/distro/tree/security/prevent_rootfs_escape_2) that still exists today. I remember you were unsure if the approach was the best way forward and called out to all maintainers for feedback (it's archived in https://github.com/orgs/python-distro/teams/maintainers/members/archived_team_posts.json for everyone with read permissions); the topic and my branch has since been ignored altogether, until your new commits here 5 days ago. 2022 is a long time ago, and I'm not interested in getting back into this topic again in 2024, myself. Thanks for your understanding, I wish this pull request the best.

HorlogeSkynet:
Hey @hartwork, and thanks for your answer. Indeed, I completely forgot about this (these) discussions, since GitHub archived them...

I took a quick glimpse of your branch which I find very complex (but this bug/feature is not trivial at all, and my new algorithm is possibly sill broken).
Although, I've updated a test case according to your commit 238925a54e91a995b923b2936c6504c262a5f638.

So I'll sum up the situation _here_ again, so @python-distro/maintainers will have the full context to properly review this (these) branches : 

---

The idea is to "fix" the `--root-dir` option (brought in python-distro/distro#161), which allows a separate rootfs to be specified (from CLI, or through `LinuxDistribution` API). As noticed first during #241, and then #311, multiple issues are related to this parameter in it's current implementation :

* [fixed during #311] this led to "unknown" binary executions, which possibly depend on host information (like the running kernel with `uname`) or even unsupported architecture (e.g. an ARM rootfs on an x86_64 host)
* paths under `root_dir` are mis-resolved to paths outside `root_dir` : this means that opening `$root_dir/etc/os-release` which links to `../../../../../etc/os-release` (or even to `/etc/os-release`) would result in host' own OS-RELEASE file being read (and distro info inferred from there)
* as a consequence, symbolic link loops are not detected (a kind of DoS could happen in some cases)

So we (@hartwork and I) worked on these issues without opting for solutions as requiring system privileges or using a third-party dependency.
There are two "final" implementations :

* The one proposed in this MR, which eventually addresses Sebastian's concerns, and which :
    1. Recursively resolves links relatively to `root_dir` until a non-link file is found (in a best-effort manner) ;
    2. Assert this file is actually under `root_dir` before returning it.

* [Sebastian's](https://github.com/python-distro/distro/commit/97edaa09e1070e030478e240d40cd7b970de3f77) (not rebased onto `master`, which manually resolves file paths relatively to `root_dir` (with full NT paths support ?)

From here, we see multiple options that I'd like to propose/discuss with you :

1. Choose and keep one of these implementations, _as is_ ;
2. Implement one of these algorithms internally, as a separate module in the `distro` package, and fully-optional (`try..except ModuleNotFoundError` for projects that vendor `distro.py` script, disabling `--root-dir` option when missing) ;
3. Implement this algorithm internally, as a separate module in the `distro` package, and required (`import [...]`) ;
4. ~~Implement this algorithm externally (a third package that _we_ [both of us] would maintain on PyPI, this would adding a [extra ?] dependency to this project)~~ ;
5. Drop the `--root-dir` parameter completely, close this MR and release distro v2 ;
6. Use a feature/trick that we do not know about, safely resolving links relatively to a specified directory (and cross platform ?).

---

Thanks for your time, see you all :wave:

HorlogeSkynet:
@adamjstewart As #369 is now closed and we eventually don't "open" another file, I take the liberty to re-propose this patch to address concerns raised in #311. I agree it's a big piece, but as it's safely guarded by `if self.root_dir is None` condition, it'd be harmless for 99,99% of cases.

adamjstewart:
@HorlogeSkynet this one's a bit outside my area of expertise. I think someone else would be better to review this, as I barely understand the problem it's trying to solve.

HorlogeSkynet:
> @HorlogeSkynet this one's a bit outside my area of expertise. I think someone else would be better to review this, as I barely understand the problem it's trying to solve.

Two years and a half, except for Sebastian, no one really gave a damn about this issue.
TL; DR : if you specify `-r /mnt/chroot` with `/mnt/chroot` being a rootfs containing, let's say, a symbolic link resolving to `/etc/os-release`, distro will parse data from your system.
Before 2a89f76cff4e6b219ae51b1cd8fe377e88558266 it was definitely a security issue as any program could be executed when distro tried `uname` or `lsb_release` data sources.
Since 2a89f76cff4e6b219ae51b1cd8fe377e88558266, it's still a (_minor_ ?) security issue as a rootfs could be misidentified as host OS, or worst : a malicious rootfs may lead to arbitrary host files parsing (e.g. `/mnt/chroot/etc/os-release` resolving to `/etc/shadow`).

This PR resolves rootfs symbolic links relatively to chroot to prevent this behavior, but is still vulnerable to TOCTOU,

HorlogeSkynet:
Hello @nir0s, could we have your review/input on this ? I'd like to incorporate this (security) fix in v1.10 release.
Thanks, bye :wave:

adamjstewart:
^ @nir0s if you're back online...

nir0s:
Will take a look promptly.

src/distro/distro.py
```diff
@@ -1165,10 +1267,14 @@ def _os_release_info(self) -> Dict[str, str]:
...
-        if os.path.isfile(self.os_release_file):
-            with open(self.os_release_file, encoding="utf-8") as release_file:
+        try:
+            with open(
+                self._resolve_path_relatively_to_chroot(self.os_release_file),
+                encoding="utf-8",
+            ) as release_file:
                 return self._parse_os_release_content(release_file)
-        return {}
+        except FileNotFoundError:
```
nir0s: This only catches `FileNotFoundError`, whereas the old code guarded with `os.path.isfile()`. If `etc/os-release` is a symlink to a **dir** (or `os_release_file` is pointed at one), `open()` raises `IsADirectoryError` instead, so it propagates and crashes. That's reachable from a crafted rootfs - exactly the threat model this PR addresses - so it's effectively an unhandled-exception. `_parse_distro_release_file()` already catches the broader `OSError`; this should do the same (or at least `(FileNotFoundError, IsADirectoryError)`).

TBH, this is the only important thing I'd address... if I had to choose one thing.

src/distro/distro.py
```diff
@@ -1157,6 +1164,101 @@ def _uname_attr(self, attribute: str) -> str:
...
 
+    @staticmethod
+    def __abs_path_join(root_path: Path, abs_path: Path) -> Path:
+        rel_path = os.path.splitdrive(abs_path)[1].lstrip(os.sep)
+        if os.altsep is not None:
+            rel_path = rel_path.lstrip(os.altsep)
+
+        return root_path / Path(rel_path)
+
+    def _resolve_path_relatively_to_chroot(self, path: str) -> Path:
```
nir0s, line 1173: The logic here is correct as far as I can tell, but it's hard to audit for security-sensitive code: ~50 lines of `relative_parts[:i]` slicing, mutate-and-restart, `for/else`, and a lexical `relative_to`. I'd prefer a single-pass input-pop / output-push segment stack (mirroring `os.path.realpath(strict=False)` with a clamped root) — it's linear and much easier to reason about. Worth considering folding that style in here.

src/distro/distro.py
```diff
@@ -796,9 +797,15 @@ def __init__(
                 self.usr_lib_dir, _OS_RELEASE_BASENAME
             )
 
+            def __isfile(path: str) -> bool:
```
nir0s: This resolves + stats the path, and then `_os_release_info` resolves + opens the same path again - two walks plus a stat→open TOCTOU window. The distro-release path uses "open and catch" instead; doing the same here would drop both the double resolution and the race.

src/distro/distro.py
```diff
@@ -1157,6 +1164,101 @@ def _uname_attr(self, attribute: str) -> str:
         """
         return self._uname_info.get(attribute, "")
 
+    @staticmethod
+    def __abs_path_join(root_path: Path, abs_path: Path) -> Path:
```
nir0s: nitpicking: the dunder prefix here (and on the nested `__isfile` above) triggers name mangling, which is irrelevant for a local helper. Would switch to a single underscore.

nir0s (commented):
Approach is good... so just a few comments.

src/distro/distro.py
nir0s: Not critical, but would be nice.

src/distro/distro.py
HorlogeSkynet: You're absolutely right, addressed :+1:

src/distro/distro.py
HorlogeSkynet: I'd prefer not getting back to work on this matter to rewrite one more time this algorithm... But feel free to change implementation with your idea and I'd be pleased to review it.

src/distro/distro.py
HorlogeSkynet: If we (only) delegate this case handling to `_os_release_info`, we cannot accomplish the "backward-compatible way of setting `self.os_release_file` attribute (added in #262). We need to know whether the path actually resolve to a file or not, to properly address `/etc` or `/usr/lib` variant.
Or I missed your point here.

src/distro/distro.py
HorlogeSkynet: Renamed.

Issue #332, opened 2022-06-03
Minor version missing for Debian >= 10


netsandbox:
For Debian  >= 10 distro can't get the minor version.

This worked for Debian <= 9 at least when you have the `lsb-release` Debian package installed with `distro.version(best=True)` because `lsb_release -a` returned the major and minor version in the `Release` key:

```
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
VERSION_CODENAME=stretch
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 9.12 (stretch)
Release:	9.12
Codename:	stretch
```

But with Debian >= 10 the `Release` key from  `lsb_release -a` only has the major version:

```
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
```

For Debian >= 10 the only source for the minor version is file `/etc/debian_version`.

The `/etc/debian_version` file exists for years, even on older Debian versions, with a stable format (major.minor, with the exception that Debian <= 6 has major.minor.patch).

So maybe you can consider to parse `/etc/debian_version` to get the minor version.

HorlogeSkynet:
Hey @netsandbox,

Whereas a small patch (see below) could bring this support to distro, one could argue that this will _break_ `version` output on Debian platforms.
From [this response](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931197#30), "Debian version" actually corresponds to what you currently get from `/etc/os-release` or `lsb_release`.
[...]

PR #333, opened 2022-06-05
Supports `debian_version` file content for `version(best=True)` calls

HorlogeSkynet:
Dear @python-distro/maintainers,

---

From ansible/ansible#77928 and #332, it appeared that Debian >= 9 do not ship anymore the `MINOR` SemVer segment in `lsb_release` "distribution release" output.
This means that setting `best=True` cannot be used to retrieve it, and thus `version(best=True)` behaves differently across Debian < 10 and Debian >= 10.

This patch implements `/etc/debian_version` file parsing to workaround this issue, on "Debian-like" distributions.

---

Many thanks, bye :wave:


tests/test_distro.py
```diff
@@ -138,11 +138,10 @@ def _setup_for_distro(self, distro_root: str) -> None:
 class TestOSRelease:
     def setup_method(self, test_method: FunctionType) -> None:
         dist = test_method.__name__.split("_")[1]
-        os_release = os.path.join(DISTROS_DIR, dist, "etc", "os-release")
         self.distro = distro.LinuxDistribution(
             include_lsb=False,
-            os_release_file=os_release,
             distro_release_file="path-to-non-existing-file",
+            root_dir=os.path.join(DISTROS_DIR, dist),
```
hartwork, line 144: The intentions (and need) for the changes to `setup_method` above are not clear to me. Are they needed? Could you elaborate?

src/distro/distro.py
```diff
@@ -898,6 +898,15 @@ def version(self, pretty: bool = False, best: bool = False) -> str:
         if self.uname_attr("id").startswith("aix"):
             # On AIX platforms, prefer oslevel command output.
             versions.insert(0, self.oslevel_info())
+        elif self.id() == "debian" or "debian" in self.like().split():
+            # On Debian-like, add debian_version file content to candidates list.
+            try:
+                with open(
+                    os.path.join(self.etc_dir, "debian_version"), encoding="ascii"
```
hartwork, line 905: This additional read^^ may interfere with #328, i.e. one of the two will need a rebase and additional changes after adjustments.

hartwork (commented):
Hi @HorlogeSkynet I like the general direction :+1:, please see two comments on details below:

HorlogeSkynet:
> Hi @HorlogeSkynet I like the general direction +1, please see two comments on details below:

Hey @hartwork ! Thanks for your quick feedback, see my replies to your comments :wink:

tests/test_distro.py
HorlogeSkynet: It appeared that `TestOSRelease` test cases **were not** "chroot-ed" in their corresponding directories... Now that we open a file under `/etc` it became an issue (host's `/etc/debian_version` was read).
I've had to set `root_dir` to workaround this as well :confused:

src/distro/distro.py
HorlogeSkynet: Definitely ! If we merge this beforehand, I'll perform a rebase for #328 :angel:

tests/test_distro.py
hartwork: The access to `self.etc_dir`, I see! :+1:

hartwork (approved):

PR #334, opened 2022-06-05
Drop support for end-of-life Python 3.6

hartwork:
https://endoflife.date/python

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,12 +28,11 @@ jobs:
     strategy:
       matrix:
         python:
-          - "3.6"
           - "3.7"
           - "3.8"
           - "3.9"
           - "3.10"
-          - "pypy-3.6"
+          - "pypy-3.7"
 
     steps:
       - uses: actions/checkout@v3
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,11 +19,11 @@ If you want to jump into the API description right away, read about the
 Compatibility
 =============
 
-The ``distro`` package is supported on Python 3.6+ and PyPy, and on any Linux
+The ``distro`` package is supported on Python 3.7+ and PyPy, and on any Linux
 or BSD distribution that provides one or more of the `data sources`_ used by
 this package.
 
-This package is tested on Python 3.6+ and PyPy, with test data that mimics the
+This package is tested on Python 3.7+ and PyPy, with test data that mimics the
 exact behavior of the data sources of `a number of Linux distributions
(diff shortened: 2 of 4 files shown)


HorlogeSkynet:
Hey again @hartwork ! Do you think dropping Python 3.6 (indeed EOL) would help us drastically reduce technical debt for this project ?
What I have in mind is [the Ansible's core team reaction](https://github.com/ansible/ansible/issues/77027#issuecomment-1040627882) when we dropped Python 3.5 lately... I fear dropping 3.6 may bother many developers vendoring distro in their package.
Tell me what you think about this !

PS : I also note maintaining multiple branches requires additional work (and time !) and we often [forget about it](https://github.com/python-distro/distro/tree/python2.7-support).

hartwork:
@HorlogeSkynet I see, I did not have those branches and Ansible on the radar. To be frank, I think it makes zero sense to put volunteer time (or even anyone's time) into maintaining end-of-life Python support — branches or not — just because some enterprises do not manage to update their stack beyond state of a museum — that's on them. If that's the constraint bubble we're in here, I'll probably be out after we finished PR #328 together.

HorlogeSkynet:
> To be frank, I think it makes zero sense to put volunteer time (or even anyone's time) into maintaining end-of-life Python support — branches or not — just because some enterprises do not manage to update their stack beyond state of a museum — that's on them.

I completely agree.

> If that's the constraint bubble we're in here [...]

I didn't intend to put their burden onto ourselves, I just meant that (IMHO) Python 3.6 **currently** doesn't constrain distro code base.
To put it another way, of course that I would have welcomed 3.6 drop if a bug, a new feature and/or one of our dependencies force us to do it (for instance, as done in #281 to allow type hints introduction).

> [...] I'll probably be out after we finished PR #328 together.

😢 the world needs you here too Sebastian

hartwork:
@HorlogeSkynet fully agree, but (for myself) I decided to not suggest drop intermediate version here because of https://github.com/python-distro/distro/blob/6d446622202849832496efd86031e5fc98ac909d/src/distro/distro.py#L628 and https://github.com/python-distro/distro/blob/bab1b9a2d6289f89e2b8c565ca3b0d3719067d43/src/distro/distro.py#L55 .

nir0s:
Can we close this? I generally agree that unless there's some overhead, there's no reason to "remove" support for Python versions.

Obviously, the views on this are very subjective. I'm with you in general (officially supporting only non-EOL versions). However, the reality is that many companies still run 2.6/7, even. I suggest that once we identify an overhead, we request that someone maintains a 3.6-support branch, just like we did with 2.7.

WDYT?

PR #339, opened 2022-06-07
Backports some patches to Python 2.7 branch

HorlogeSkynet:
PS : I had to adapt #331 due Python 2.7 support drop by upstream `black` and underlying `click` dependency...


nir0s (approved):
Overall LGTM. I admit that I'm not following up on each and every change, though.

HorlogeSkynet:
> Overall LGTM. I admit that I'm not following up on each and every change, though.

I've "only" cherry-picked commits that could _easily_ be applied, the other ones are left for those who would need them.
Thanks for reviewing this :bow:

PR #355, opened 2023-02-20
Run Python 3.6 on Ubuntu 20.04 for CI and bump isort

HorlogeSkynet:


.github/workflows/ci.yaml
```diff
@@ -30,12 +30,15 @@ jobs:
...
           - "3.7"
           - "3.8"
           - "3.9"
           - "3.10"
-          - "pypy-3.6"
+        include:
+          - os: "ubuntu-20.04"
+            python-version: "3.6"
+          - os: "ubuntu-20.04"
+            python-version: "pypy-3.6"
```
hartwork, line 41: The values for `os` are not really used anywhere, `runs-on` further up does not respect the matrix. Am I missing something?

hartwork (changes requested):

.github/workflows/ci.yaml
HorlogeSkynet: Good catch, I had actually back-ported a patch on a different matrix. 3.6 jobs are now included !

hartwork (approved):
@HorlogeSkynet :+1:

Issue #322, opened 2022-01-12
uname_info() return empty string on ubuntu20.04 and Fedora34


braindevices:
is this expected?
```
distro.__version__
 '1.6.0'

print(distro.uname_info())
{}
```

sethmlarson:
Maybe? What does running `uname -a` print on those platforms?

braindevices:
the uname installed fine.

```
distro.LinuxDistribution(include_uname=True)
LinuxDistribution(os_release_file='/etc/os-release', distro_release_file='', include_lsb=True, include_uname=True, _os_release_info={'name': 'Ubuntu', 'version': '20.04.4 LTS (Focal Fossa)', 'id': 'ubuntu', 'id_like': 'debian', 'pretty_name': 'Ubuntu 20.04.4 LTS', 'version_id': '20.04', 'home_url': 'https://www.ubuntu.com/', 'support_url': 'https://help.ubuntu.com/', 'bug_report_url': 'https://bugs.launchpad.net/ubuntu/', 'privacy_policy_url': 'https://www.ubuntu.com/legal/terms-and-policies/privacy-policy', 'version_codename': 'focal', 'ubuntu_codename': 'focal', 'codename': 'focal'}, _lsb_release_info={'distributor_id': 'Ubuntu', 'description': 'Ubuntu 20.04.4 LTS', 'release': '20.04', 'codename': 'focal'}, _distro_release_info={}, _uname_info={})

subprocess.check_output(('uname', '-rs'))
b'Linux 5.4.0-109-generic\n'
```
Any idea what happens here?

braindevices:
I finally have time to dig into this. It turns out the [uname content parser returns empty dict by design](https://github.com/python-distro/distro/blob/b64ce77605683e28a994cd6c6f5a2d1f0886945e/src/distro/distro.py#L1222). So the `uname_info()` is actually never supposed to be called outside. It is merely a fallback method to get the distro id and distro version for some os such as AIX or some mac os. And even for those os the uname_info() does not really return the full kernel version.

So I think we should NOT expose the uname_info() and uname_attr() to user.

HorlogeSkynet:
Hello @braindevices, indeed, when `uname -rs` gives `Linux [...]` the release part is simply ignored. Let's deprecate direct usages of `uname_info` and `uname_attr` to prepare a switch to protected methods in distro v2 :+1:

PR #370, opened 2024-08-22
Deprecates `uname_attr` and `uname_info` public methods

HorlogeSkynet:
(see #322 for context)

distro as well as LinuxDistribution `uname_attr` and `uname_info` public methods are based on `_parse_uname_content` function which purposely ignores release information part from `uname -rs` command output on Linux platforms. This makes it specially designed for distro internals, and shouldn't be publicly available as stable API.

We'll deprecate these methods in v1.11.0, in order to allow API removals in the future (e.g. distro v2).

### DIFF
--- a/src/distro/distro.py
+++ b/src/distro/distro.py
@@ -546,9 +546,19 @@ def distro_release_info() -> Dict[str, str]:
 
 def uname_info() -> Dict[str, str]:
     """
+    .. deprecated:: 1.11.0
+
+        :func:`distro.uname_info()` is deprecated and will be removed in a
+        future version.
+
     Return a dictionary containing key-value pairs for the information items
     from the distro release file data source of the current OS distribution.
     """
+    warnings.warn(
+        "distro.uname_info() is deprecated and will be removed in a future version.",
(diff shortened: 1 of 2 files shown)


HorlogeSkynet:
Up @python-distro/maintainers 🙏

HorlogeSkynet:
> Sounds good to me.

Thanks for your feedback !
Amended and added pytest configuration to ignore our own deprecation warnings not to pollute tests execution logs.

adamjstewart (approved):
Could also remove it from the documentation. I don't know much about these tools but trust you to make the decision of whether or not to deprecate. I'm just approving the implementation.

HorlogeSkynet:
> Could also remove it from the documentation.

Let's not while they actually exist (mostly because one might use them on the v1 branch).

> I don't know much about these tools but trust you to make the decision of whether or not to deprecate. I'm just approving the implementation.

Thanks ! These API are misleading users, as brought up by @braindevices in #322. I'm merging this to move on towards [v1.10](https://github.com/python-distro/distro/milestone/6).

PR #374, opened 2024-12-21
Ignore "n/a" from `lsb_release` output allowing other sources to be honored

### LINKED ISSUE #373: distro.version() returns 'n/a' in Debian13
distro.version() is supposed to eliminate `n/a` and empty values when other sources contain valid values.

```
>>> import distro
>>> distro.version()
'n/a'
```

It returns `n/a` because `lsb_release -a` has that value, even though `/etc/debian_version` contains a valid value
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux trixie/sid
Release:	n/a
Codename:	trixie

```
```
$ cat /etc/debian_version 
trixie/sid
```

HorlogeSkynet:
> closes #373


HorlogeSkynet:
Up @python-distro/maintainers please :pray:

adamjstewart (approved):

